PR #14, opened 2015-03-09
make nana compiling with mingw 4.9.2

Pr0curo:
Hi Jinhao,

this little patch is to make sure nana can compile with the current mingw 4.9.2.
It is tested to compile with mingw and msvc12 (2013) aswell.

I think the changes are self-explanatory. please feel free to ask me anything about the changes.

best regards, Jan

ps. if something is not right with the git usage pls tell as i'm fairly new to git(hub)


cnjinhao:
Thanks for you patch!
It would be better if you commit a pull request into a develop branch or hotfix branch.

cnjinhao:
The strings functions can make the library work on MinGw 4.8.1, IMO, we should not delete them now. They will may be deleted in a certain future release.

cnjinhao:
This commit has been manually merged into hotfix branch. Thank you!

PR #16, opened 2015-03-13
doxy-comments and place-split avoiding 0 left size

qPCR4vir:
I´m not sure, but It seems that the new leaft_left_ size is calculated by scaling the previous value. But if the first size was 0, no matter how big the scale is, it will be 0 always.

### DIFF
--- a/include/nana/gui/widgets/button.hpp
+++ b/include/nana/gui/widgets/button.hpp
@@ -1,13 +1,13 @@
-/*
+/**
  *	A Button Implementation
  *	Nana C++ Library(http://www.nanapro.org)
- *	Copyright(C) 2003-2014 Jinhao(cnjinhao@hotmail.com)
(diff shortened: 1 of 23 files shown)


cnjinhao:
>But if the first size was 0, no matter how big the scale is, it will be 0 always.
How to reproduce this issue?

qPCR4vir:
Hi, I think I solved the problem for now with another workaround in my code. 
But anyway, I have reproduced the issue. You can see it in:
https://github.com/qPCR4vir/nana-demo/blob/master/Examples/a_group_impl.cpp
This is with the uncomment // problem    line:
group
and commenting that line and uncomenting the // OK line:
group-ok
The add butoon is correctly shown in 30% left to the others two buttons.

cnjinhao:
Got it, It seems a possible solution.

PR #17, opened 2015-03-17
events doxy comments

qPCR4vir:
Hi,
two Q:
arg_dropfiles could be derived from arg_mouse ?
It is possible to have something like: struct arg_drop:arg_mouse {any data;}; ?

### DIFF
--- a/include/nana/gui/detail/general_events.hpp
+++ b/include/nana/gui/detail/general_events.hpp
@@ -430,8 +430,7 @@ namespace nana
 		unsigned distance;	///< expressed in multiples or divisions of 120
 	};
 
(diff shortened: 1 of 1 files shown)


cnjinhao:
arg_mouse contains some data which are not included in arg_dropfiles, and WM_XBUTTONUP does not occur when WM_DROPFILES occurs, you never know which button user will release. So it's a bit more complicated to maintain these data.

qPCR4vir:
Oh, I see, complicate. OK. One need to asume: use these files, and better not: move these files (becouse we dont know about shift/cntrl). That is becouse the event originated in another application, probable Windows Explorer? OK. I assume than the other Q is complicated too.

PR #18, opened 2015-03-17
Use bakefile for makefile generation

kirbyfan64:
I described this a bit ago. Instead of hand-writing makefiles and VS 2013 files, they are now auto-generated via [Bakefile](http://bakefile.org/).

Pros:
- No need to maintain two different sets of build files.
- The generated makefiles are REALLY fancy, supporting things like debug/release builds..

Cons:
- You need to run `bkl nana.bkl` in `build/bakefile` whenever you want to regenerate the build files.

I think the pros outweigh the cons. Of course, feel free to reject this if you want.

Bakefile is ridiculously easy, though, so it's not like learning a crappy macro language (*cough* CMake *cough*).


qPCR4vir:
Hi,
this looks OK. But the phrase: 
This file was generated by Bakefile (http://bakefile.org). Do not modify, all changes will be overwritten! 
Is very intimidating. Many of us are not ready for that. I think that a good solution will be to change the .bkl file to:

gnu.makefile = ../build/makefile-bkl/makefile; 
vs2013.solutionfile = ../build/vc2013-bkl/nana.sln; 

And this automatically generated files will not be included in the official nana repository, but will be generated by the end user in directories ignored by git.I think this way all of us will be happy and will try this new tool without fear.

kirbyfan64:
Well, the basic idea behind Bakefile is to check in the files into version control...but I can do that if you really want it.

qPCR4vir:
Yes, I see. The Problem with puting the generated files in the repertory is only not to "bloat" it. But that force each to run self the bkl... so you are right it is best to have the projects/makes files equal and ready for all. 
But ... hmm. Ok. And how to create the folders (filters?) in the "Source Files" section(.cpp)? And in the "Header Files" section (.h) of the IDE Project (VS2013)? 
Oh good, it generate the ../../inlude. How to set the sources dir for IntelliSense?
Of course.. if we will master that, it will be great to generate the "user's" projects, for example the ones to compile the examples using the nana.lib :-). 
hmm, I see (reading the docs), we need:
```archs = x86 x86_64;```
```win32-crt-linkage = “static” or “dll” ```
etc.

cnjinhao:
Agree with qPCR4vir, ignoring the automatically generated files and generating these files in a separate directory are more friendly to the user.

Thank you for providing a bakefile solution!

kirbyfan64:
@cnjinhao Done.

cnjinhao:
Thanks for your contribution!

PR #28, opened 2015-04-02
fixing lookup error from gcc/mingw

Pr0curo:
small change to make nana compile with mingw again.

best regards, jan
^


cnjinhao:
Thank you for your contribution

PR #29, opened 2015-04-05
enable MultiProcessorCompilation to decrease build time

beru:
Please enable MultiProcessorCompilation to decrease build time.


cnjinhao:
Thanks for your contribution!

PR #31, opened 2015-04-05
fix mingw compilation error

qqiangwu:
1. The code failed to compile in mingw. I fixed it.
2. Add [biicode](https://www.biicode.com/) support.


cnjinhao:
Thanks for your patch!

PR #33, opened 2015-04-08
fix wrong positions of menu and tooltip in secondary monitor

qPCR4vir:
Hi,
How are you?
You were in the right direction.
This will probably fix other errors. 
Now is OK with the secondary monitor.


cnjinhao:
Thanks for your fix!

PR #35, opened 2015-04-11
modify readme and detect environment for biicode automatically

qqiangwu:
1. Add Biicode example in README
2. Eliminate macro redefinition warnings when used with Biicode. When used with Biicode, the default config.hpp will be used which will automatically detect its environment. When used individually(via raw CMake), nothing changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ Thumbs.db
 #Ignore files build by Visual Studio
 bii/build/*
 bii/cmake/*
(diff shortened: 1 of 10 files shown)


cnjinhao:
Thank you for your contribution, but I deleted the marco NANA_NO_CPP11, IMO, it's not a good idea. The STD_THREAD_NOT_SUPPORTED in hotfixes branch is automatically detected

PR #40, opened 2015-04-20
Feat page up dw listbox

qPCR4vir:
This changes are also avoailables from [my master](https://github.com/qPCR4vir/nana/tree/master)
Comented in https://github.com/beru/nana/issues/12


qPCR4vir:
This can be also inspected and tested in a new experimental dev branch, and after your main dev branch will be actualized I can do a new, clean pull req.
Actually I wanted to unify the functions to scroll the listbox, and then reuse them from the keyboard events and from events from the scrollbar, but until now I didtn fond how. 
I will try to fix the incorrect order of scrolling with the key up and down when the list is ordered. With the scrollbar it scroll always OK. Also, when the shift key is pressed page up and down need to expand the selection, not reset it.

qPCR4vir:
I continued this in my branch [page_scroll](https://github.com/qPCR4vir/nana/tree/page_scroll) using the scroll as I wanted.

qPCR4vir:
A new pull https://github.com/cnjinhao/nana/pull/43 is better

PR #44, opened 2015-04-24
FIX / FEATURE: Listbox Export/copy CSV-like; handle Home, End. OK in windows.

qPCR4vir:
From https://github.com/beru/nana/issues/12#issuecomment-95764619
Sorry, I cant test in Linux.


qPCR4vir:
Re - tested after merge. Its seems OK!

qPCR4vir:
incorrect comment:

> FIX: a crash when clear an empty listbox incorrect comment:

correct comment:

> FIX: a crash when user press home, up, page-up, etc. keys on an empty listbox (impossible to select an item)

cnjinhao:
Thank you for your effort!

PR #48, opened 2015-04-26
add nana::system::dataexch::set(const nana::paint::graphics& g) method

beru:
Hi, I pushed code to set bitmap into clipboard. 
It only supports Windows platform though.


cnjinhao:
Thank you for your contribution

PR #49, opened 2015-04-26
add methods to get/set sortability of listbox

beru:
There are some cases I want to disable sorting feature of listbox.

### DIFF
--- a/include/nana/gui/widgets/listbox.hpp
+++ b/include/nana/gui/widgets/listbox.hpp
@@ -552,7 +552,10 @@ By \a clicking on a header the list get \a reordered, first up, and then down al
 			_m_ease_key(&key);
 		}
 
-		            ///Sets a strict weak ordering comparer for a column
+		bool sortable() const;
+		void sortable(bool enable);
+		
+		///Sets a strict weak ordering comparer for a column
 		void set_sort_compare(size_type col, std::function<bool(const nana::string&, nana::any*,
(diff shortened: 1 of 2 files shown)


qPCR4vir:
Hi,
I tested it, and works. 
It prevent to change the sort column from the user interface. 
it don't change whatever was before and you can still programmatically set a sorting column.
Ok.
I agree that some time we can need a more "static" list and these function helps customizing that. Another can be: fixed_width(bool) to prevent resizing columns, or no_copy, no_export.

Oh, revisiting this I sow we need to return from functions useful during initialization a ref to this listbox, to allow chaining of initializations, just like header_width() and orderer_categories().

And in the implementation of enable_single I see internal_scope_guard lock; We need to use it in most of the implementation of the other functions?

beru:
Hi, qPCR4vir. Thanks for the review.
As you have mentioned, this fix is incomplete and has some flows.
I only cared about sorting by user interaction.

I'm not sure how much of thread safety has been in mind regarding this library.
Personally, I do not think the library should handle all the corner cases occur with multi-threaded GUI software. Applications should have some responsibility before giving procedures to the library.

cnjinhao:
The internal_scope_guard is a lock to protect the internal data structure from multi-threaded read/write operations. The API::functions are always thread-safe, they are protected by the internal_scope_guard, but other functions are not.

PR #51, opened 2015-05-02
Fix Clang builds (again)

kirbyfan64:
This fixes the following:

- An ambiguous call to `std::abs` (of course, it's not really ambiguous; glibc is just broken)
- `algorithm` wasn't included, so the calls weren't getting resolved correctly


kirbyfan64:
Actually, don't merge this yet. I just filled my screen with many more build errors compiling [Hello, world!](https://github.com/qPCR4vir/nana-docs/wiki/Hello-World).

kirbyfan64:
Nevermind. False alarm. These are the only fixes needed.

cnjinhao:
Fixed

PR #53, opened 2015-05-03
remove some warnings with VS2013

beru:
Hi, I edited nana's code a little bit to remove some warnings with VS2013.
And in time::time(unsigned hour, unsigned minute, unsigned second) constructor, I added return statement if args are valid.


cnjinhao:
Nice improvement!

PR #54, opened 2015-05-05
Hotfixes 1.0.2

qPCR4vir:
I was testing for first time listbox with many categories. See example in:
https://github.com/qPCR4vir/nana-demo/blob/master/widget_show2.cpp


cnjinhao:
These fixes have already merged into hot fixes branch

PR #55, opened 2015-05-10
Letting cmake to decide which c++ standard to use

rajko-d:
If compiler doesn't support c++14 it will fall back to c++11 as was before. It compiles without warnings on win 8, mingw gcc 5.1 but I understand that doesn't mean anything."


cnjinhao:
Thanks for your contribution, I am looking forward to your next pull request.

PR #68, opened 2015-07-19
Develop group

qPCR4vir:
Hi,
(sorry for the many merge commit.. ?)
Thank, you have got raid of the secound panel in group and it make the interfase much better.
I adapted the samples:
https://github.com/qPCR4vir/nana-demo/blob/master/Examples/group.cpp

I also added a group constructor. You know, I like to construct in one step when possible.
I fixed two small problems.

I also compiled and run all the examples. There are some problems.
1- crash at begging:
FreeMe_v140_Debug_x86
draw_v140_Debug_x86

2- don't work and exception crash: graphics is empty
stretch_image_v140_Debug_x86

3- crash by closing the nana window, but not the system (console) windows:
widget_show2_v140_Debug_x86   --> but Now it shows the image in the button
widget_show_v140_Debug_x86    --> but Now it shows the image in the button
a_group_impl_v140_Debug_x86

Something with image?


cnjinhao:
The image::paste/stretch now throw an exception if the image object or graphic object is empty. It is good for avoiding potential bugs, but, as you can see, it may be a bit complicated, the graphics should be checked before calling image::paste/stretch.

I think another solution is much better.
```
void image::paste(const rectangle& r_src, graphics & dst, const point& p_dst) const
{
	if (image_ptr_)
	{
		if (dst.empty()) //create if it is empty, instead of throwing a exception
			dst.make({ r_src.width, r_src.height });	//throws if failed to create

		image_ptr_->paste(r_src, dst, p_dst);
	}
	else
		throw std::runtime_error("image is empty");
}
```

PR #72, opened 2015-08-22
Fix missing array declaration in a unique_ptr

dareg:
Which was causing a wrong deletion of the array


dareg:
And by the way, thanks for your library ! :)

PR #73, opened 2015-09-13
Fix Visual Studio 2015 codecvt linker errors

roberthauck:
Add missing defines


cnjinhao:
Thank you for this patch

PR #85, opened 2015-10-27
return configuration specific names

qPCR4vir:
To avoid excessive recompilation after each change in configuration and conflicts during linking.
It allows simple selection of configuration inside the IDE.

This is optional, just convenience.


qPCR4vir:
Explanation here:

https://github.com/qPCR4vir/nana-docs/wiki/Thoughts-about-configurations-and-files-names-in-VS

PR #86, opened 2015-11-06
Fix listbox for auto_draw+inline_factory+item_proxy::value.

leobackes:
When calling templated append() overload on a shown listbox, auto_draw
caused an early redraw, which issued the inline widget set() member
function before the value() and value_ptr() were actually set.


cnjinhao:
Sorry, that is my mistake, the cat_proxy::update() is private. I renamed it _m_update

PR #91, opened 2015-11-24
Clean-up build

qPCR4vir:
We all see that most of the questions are about building. We normally have answers for all of that, because Nana support many options, which is great. I can not help very much, because I use MSVC only. But I think that a little clean-up can help a lot. 
I have tried this in the wiki. [Please take a look: is all right there](https://github.com/qPCR4vir/nana-docs/wiki/Installation) ?


cnjinhao:
Hi, the external library configuration has been changed since version 1.2. It's necessary to remove the content of extrlib from installation instructions step one, adding the new [configuration instruction of extrlib](https://github.com/cnjinhao/nana/wiki/Configuration-of-Third-Party-Libraries-for-Nana) will be of great help.

PR #93, opened 2015-12-03
Cmake mingw

qPCR4vir:
Hi !
I'm sorry. This maybe better.

I installed CLion and CMake (it is good, but the best is that one don't need admin password to install it, the same with MinGW) with MinGW and make. I think I clean-up old things all around, making it more consistent. I concentrated all the info into config.h, which one can also configure manually or automatically with CMakelist. I managed to compile libnana.a - no errors.

If something is not clear please ask...


cnjinhao:
Significant improvements. Thank you!

cnjinhao:
I removed the check of NANA_WINDOW, NANA_LINUX and NANA_MACOS at the beginning of platform_spec.cpp, because these flags may not be defined at that point.

qPCR4vir:
I see. You are right.
I propose to use the check but not at the beginning of the file `platform_spec.cpp` but rigth after 

```
#include <nana/detail/platform_spec_selector.hpp>
#if defined(NANA_WINDOWS)
```

this way we can eliminate the `platform_spec_selector.cpp`

Another less conventional solution, but very simple is to move both `platform_spec.cpp` files to the include directory, together with the `platform_spec.hpp` files and keep the `platform_spec_selector.cpp`. This way the automatic project generator will not include the cpp files in the build. Of course, only if the tool scan only the source dir and not the whole nana base dir with the include dir.

cnjinhao:
Yes, I did not take account automatic generator. BTW, I updated the config.hpp for detection of flags on GCC.

PR #94, opened 2015-12-04
Fix PLATFORM_SPEC_HPP defines

chekoopa:
They're just missing from config.hpp, unlike CMakeLists, and that prevents any manual builds (e.g. using Code::Blocks).

### DIFF
--- a/include/nana/config.hpp
+++ b/include/nana/config.hpp
@@ -36,7 +36,8 @@
 #if defined(_WIN32) || defined(__WIN32__) || defined(WIN32)
 //Windows:
 	#define NANA_WINDOWS	1
-
+	#define PLATFORM_SPEC_HPP <nana/detail/win32/platform_spec.hpp>
+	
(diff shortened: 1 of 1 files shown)


cnjinhao:
Thank you for fixing this issue. In fact, the PLATFORM_SPEC_HPP is removed from 1.2, but the released package is totally wrong, the version of source files is not 1.2.

qPCR4vir:
Ohh, the same problem:

Hi !
Thank you for your feedback. It is very usefull.
Jinhao is right now working with the configuration.
Could you please try the very last commit at:

https://github.com/cnjinhao/nana/tree/hotfix-1.2

? Are you using git? If not, just donwload there the ziped source code.
The PLATFORM_SPEC_HPP is not more used, and there are many other small related changes.
Thank!

http://nanapro.org/en-us/thread.php?id=07dfc996111e86c701d65833a9e202e0&stmp=1449322122083

PR #97, opened 2015-12-24
Adding double click event on items, allows block expansion

leobackes:
Adds a new event in listbox to notify when the user double-clicks an item.

### DIFF
--- a/include/nana/gui/widgets/listbox.hpp
+++ b/include/nana/gui/widgets/listbox.hpp
@@ -474,15 +474,14 @@ namespace nana
 
         arg_listbox(const drawerbase::listbox::item_proxy&, bool selected);
     };
-	struct arg_item
+    struct arg_category
 		: public event_arg
 	{
-        drawerbase::listbox::index_pair item;
-        drawerbase::listbox::size_type column;
+        drawerbase::listbox::cat_proxy category;
         void block_category_change() const;
         bool category_change_blocked() const;
 
-        arg_item(const drawerbase::listbox::index_pair&, drawerbase::listbox::size_type column);
+        arg_category(const drawerbase::listbox::cat_proxy&);
     private:
         mutable bool _m_block_change;
 	};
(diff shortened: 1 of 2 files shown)


include/nana/gui/widgets/listbox.hpp
```diff
@@ -461,13 +461,25 @@ namespace nana
...
 		: public event_arg
 	{
-		mutable drawerbase::listbox::item_proxy item;
-		bool	selected;
-
-		arg_listbox(const drawerbase::listbox::item_proxy&, bool selected);
+        drawerbase::listbox::index_pair item;
+        drawerbase::listbox::size_type column;
+        void block_category_change() const;
+        bool category_change_blocked() const;
```
cnjinhao: Thanks for your pull request, I think it would be much better if this event only occurs when double click on a category to determine whether to change the status of expansion of a category. Because block_category_change() and category_change_blocked() may confuse us when clicking on an item(not category).

By using general double click event, it is able to handle the double click of an item.
```C++
list.events().dbl_click([&list](const arg_mouse& arg)
{
	auto s = list.selected();
	if (s.empty())
	{
		//auto col = list.column(arg.pos); //Add a method to get the column index with a specified coordinate
	}
});
```

leobackes:
Hey, thanks for accepting it :)

When coding this I thought of having either item_dbl_click and category_dbl_click.
The main point of this separate event is because dbl_click + selected() is eventually unreliable, as it (before) would have all items of a category selected upon double-clicking it.
Also, dbl_click is issued when double-click happens anywhere in the control, so selecting an item and double-clicking a header would also trigger the "item double click" in this case.
It is not possible to know in which of the columns the event has happened;

Maybe we could make possible to the user-code to check if the click position is contained in the area of the item?!

Btw, I couldn't think of a better name, but arg_item seems too generic to belog to nana namespace...

cnjinhao:
>>Maybe we could make possible to the user-code to check if the click position is contained in the area of the item?!

Yes, the function to check a position is also useful to detect an item in mouse_move event. Therefore, the event for item can be eliminated.

leobackes:
Hi,

Made some updates.
Added overload of at() function taking a nana::point and returning an index_pair;
Added column_from_pos() also taking a point, it returns the index of the column at the given pos. Maybe this could take just an size_type for the x position, as the y wouldn't matter, but this could be confusing.

Changed the event to an arg_category, and it will now be only emitted when dbl clicking a category.

cnjinhao:
Thanks for your contribution!

PR #98, opened 2016-01-03
Fix for mingw users (incomplete)

Pr0curo:
here are three little fixes for the current 1.3 alpha release.

hope the commits are self-explanatory. if not pls give me ring.

1. Changed CMakeLists.txt encoding to utf-8 (i simply suppose thats how it should be)
2. CMake 3.3.0 told me there was a problem with 'if' --> see the change 
3. seems like in Microsofts STL already includes ```<cstring>``` at some point, but with libstdc++ it was required to include it explicitly

there is another issue i could not fix:
using Microsofts STL it is possible to open a std::basic_ifstream<char> with an wchar_t* path (even though it's not standard behaviour)

```
Error:

In file included from D:\nana\source\paint\image.cpp:31:0:
D:\nana\source\paint\detail/image_bmp.hpp: In member function 'virtual bool nana::paint::detail::image_bmp::open(const nana::experimental::filesystem::path&)':
D:\nana\source\paint\detail/image_bmp.hpp:88:57: error: no matching function for call to 'std::basic_ifstream<char>::basic_ifstream(const value_type*, const openmode&)'
     std::ifstream ifs(filename.c_str(), std::ios::binary);
```

nana makes use of this non-standard-feature in the image_bmp.hpp.

Used compilers:
MSVC2015 --> working
MinGW gcc-5.1.0-tdm64-1-c++ --> errors as described

best regards,
Jan

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@
 
 #https://cmake.org/cmake-tutorial/
 #https://cmake.org/cmake/help/v3.3/module/CMakeDependentOption.html?highlight=cmakedependentoption
-# use CACHE FORCE  or set(USE_meganz_mingw_std_threads ON) or delete CMakecache.txt or the entirely build dir
+# use CACHEÂ FORCE  or set(USE_meganz_mingw_std_threads ON) or delete CMakecache.txt or the entirely build dir
 # if your changes don't execute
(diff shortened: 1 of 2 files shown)


cnjinhao:
Thanks for your feedback.

PR #102, opened 2016-01-23
Fix for menu images

kmribti:
Stretch menu image (icons) only when don't fit, center it otherwise.


cnjinhao:
Thanks for your patch.

PR #104, opened 2016-01-24
Toolbar improvement.

kmribti:
This patch has 2 new features:

 1- new option at constructor to enable creation of detached toolbar from his parent, this enable the use of toolbar anywhere and the size can be controlled by a layout manager.

 2- add a new option in order to enable to place buttons at right part, the widget has a new method go_right() after calling it, every new button is right aligned.

Both changes don't break compatibility with old code.

### DIFF
--- a/include/nana/gui/widgets/toolbar.hpp
+++ b/include/nana/gui/widgets/toolbar.hpp
@@ -92,6 +92,13 @@ namespace nana
 		bool enable(size_type index) const;
 		void enable(size_type index, bool enable_state);
 		void scale(unsigned s);   ///< Sets the scale of control button.
+
+		void go_right();
+
+		bool detached() { return detached_; };
+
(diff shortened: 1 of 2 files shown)


kmribti:
I attach a screenshot with an example: 

toolbar

qPCR4vir:
Hi @kmribti,
your screenshot look really very good.
It seems that already have matured the possibility to together build a super widget tree_list or tree_explorer composed by:
menu bar (optional)
toolbar (optional)
categorice (optional)
tree (optionally with more than one base node)
list 
item view (optional, in your case a textbox, bot in general a group) with an optional navigator tool (!)
status_line (optional)

You have this. Jinhao have the filebox for linux and @k1988 me have adapted as independent file explorer. I used also one more variant in my soft.

PR #106, opened 2016-01-27
Fix different exception specifiers in image.cpp

nabijaczleweli:
As discussed [here](http://chat.stackoverflow.com/transcript/message/28394446#28394446)


cnjinhao:
Thanks for your PR

PR #107, opened 2016-01-27
Add TravisCI support

nabijaczleweli:
As discussed [here](http://chat.stackoverflow.com/transcript/message/28394615#28394615).

Also adds a TravisCI build badge to the README.

How to activate:
  * Before merging the PR:
    - go to https://travis-ci.org,
    - log in (if necessary),
    - click your name and avatar in the top-right corner,
    - check (click on, so it turns from grey to green) `cnjinhao/nana`,
    - click on the little gear icon, which will take you to the settings page,
    - check the option in the top-left corner ("Build only if .travis.yml is present").
  * Merge the PR.

Now, each time a commit is pushed to a branch containing a `.travis.yml` file, a build will start and, if it fails, responsible individuals will be notified via e-mail.
Also, if a PR is opened (and is auto-mergable), a build will start targeting the merged state.

PR #109, opened 2016-02-12
Travis

qPCR4vir:
See #107 
Please, go to https://ci.appveyor.com/login and add this project there.


qPCR4vir:
TODO:
- activate clang in travis CI
- activate MacOS  in travis CI
- move demos - cmake code to the demo repo and here only include it. Include all demos, maybe "automatically".
- PNG ON: see the error here: https://travis-ci.org/qPCR4vir/nana/jobs/108676699
- Audio ON: [Could not find a package configuration file provided by "asound" ](https://travis-ci.org/qPCR4vir/nana/jobs/108672508)
- include demos in AppVeyor CI build.
- make "real" test: run the demos (not just compile) in CI and return if run OK.
- create a "framework" in Nana for automatic GUI testing !! By conditional compilation include a routine run in parallel (another thread) that will timely send both low and high level events (bottom click, mouse move, etc)

qPCR4vir:
Hi !
Actually the demos run OK in windows, with VC2015 and with gcc++5 mingw in Clion.
It seems like there are  errors in the `std::move` implementation used in the build in travis.
See the details, from `tabbar `trying to move an `string`. I don't see any error, and the message point to an error in what `move `return

qPCR4vir:
Urra ! Fixed !!

cnjinhao:
I reviewed early today, and found the to_nstring issue in tabbar. But now it have been fixed by your last commit. Thank you

PR #110, opened 2016-02-15
Makefile remake; vim-modeline; fencoding stuff

sarrow104:
clang? I'll try that. But I don't have MacOS.


qPCR4vir:
Hi !
Thank you. Our makefile need update. 
Maybe you could make clang to compile? MacOS? 
**Adding** make to our travis will be also good.

I agree that a consistent tab/spaces style is good, but I suggest you to separate in two differents pull request: the make-related changes in one and the spaces in the other. 
Or better, just suggest the spaces change in an "issue". It is hard to follow were are real changes and where just spaces. Also, all my git client are hanging. 
I gest nana have many hand made formating using spaces/tabs that jinhao will need to revise - that will be a separate task that he could want to think about separately.

cnjinhao:
IMO, replacing all tabs that indent the start of a line with spaces is not a good idea.

cnjinhao:
Close the PR, but makefile is accepted

PR #118, opened 2016-03-03
Automatic GUI test pass in travis

qPCR4vir:
[It is working good](https://travis-ci.org/qPCR4vir/nana/jobs/113431787).

We need to set this automatically in cmake. But first we need a general way to exit from an application from the parallel GUI testing function. Now I use `fm.close()`, and I need to pass a pointer to the main form to that function to do that. A general `API::exit()` will be much better but don't works. Why? With `API::exit()` nana can add a default GUI test to any program when compiled with that option, with is ideal for travis.  

Also, some demos compiled [but don't pass the run test](https://travis-ci.org/qPCR4vir/nana/jobs/113411795) - the list is at the end of `.travis.yml`. We need to fix that. It maybe a problem with [the Xvfb, an X server to emulate display](http://www.xfree86.org/4.0.1/Xvfb.1.html)? See hier [how to use in travis](https://docs.travis-ci.com/user/gui-and-headless-browsers/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ build/makefile-bkl/*
 *.ilk
 *.log
 [Bb]in
+[Bb]in/
 [Dd]ebug*/
 *.lib
 *.a
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,9 @@ install:
   - /tmp/tools/cmake --prefix="$HOME" --exclude-subdir
 
 before_script :
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3 # give xvfb some time to start
   - mkdir bld
   - cd bld
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,7 +128,7 @@ script:
   #- ./screen  
   - ./stretch_image  
   - ./threading  
-  - ./thread-pool  
+  #- ./thread-pool  
   - ./various_events  
   #- ./window-dragger
   - ./windows-subclassing   
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ matrix:
             - llvm-toolchain-precise 
 
 before_install:
-  - git clone --depth=50 --branch=testing https://github.com/qPCR4vir/nana-demo.git nana-demo
+  - git clone --depth=1 --branch=dev_nana_in_examples https://github.com/qPCR4vir/nana-demo.git nana-demo
   - export PATH="$HOME/bin:$PATH"
   - mkdir ~/bin
   - wget --no-check-certificate --no-clobber -O /tmp/tools/cmake https://cmake.org/files/v3.4/cmake-3.4.0-rc3-Linux-x86_64.sh || true
--- a/.travis.yml
+++ b/.travis.yml
@@ -127,10 +127,10 @@ script:
   - ./png  
   #- ./screen  
   - ./stretch_image  
-  #- ./threading  # Press a char: 
+  - ./threading  
   - ./thread-pool  
   - ./various_events  
-  - ./window-dragger
+  #- ./window-dragger
   - ./windows-subclassing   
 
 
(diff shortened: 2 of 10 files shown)


qPCR4vir:
The function exit_all exit from all threads... OK? not sure, but it works...

cnjinhao:
Hi, qPCR4vir. Is it necessary to include the auto GUI testing in release package?

qPCR4vir:
Hi,
good question. 
In the future, surely yes.
For beginners it is not a problem because it is `off `by default.
For mid-level programmers it is a big plus: you simple fork `nana` and `nana-demo`, add your own experimental app to your `nana-demo` (and in cmake and `.travis.yml`, just like all the other demos) and you will get testing in Linux for free (in the future in windows too, and possible in OSX with clang too). To me it is a very good and easy way to test that my app is really "crossplatform" and not only in theory (I can't install all of this locally just to test my app).  
But I agree that right now it is immature - it need more work. I propose you to forgot hotfix-1.3 and create [`hotfix-1.3.1`](https://github.com/qPCR4vir/nana/tree/hotfix-1.3.1) in the commit before your last merge. It have not gui testing, and could be the next release, or after merging your `develop` could be release 1.4
The GUI testing can be in a new feature branch [`Automatic_GUI_testing`](https://github.com/qPCR4vir/nana/tree/Automatic_GUI_testing). I have all of this in my [nana fork](https://github.com/qPCR4vir/nana) take a look.

qPCR4vir:
see : 
https://travis-ci.org/qPCR4vir/nana/branches

cnjinhao:
API::exit() only closes windows which are created by the calling thread.
I found that a new form pointer parameter for exec(), is the form like a main form? Will all windows get closed when the main form is closed? IMO, the idea of main form is not good.

I think the best way to do test is to add a test.cpp and define a specific function for testing rather than modify exec().

qPCR4vir:
Your are right. The main form pointer was a workaround and is now eliminated.
And yes, the workaround was needed because `API::exit()` don't close all forms.

Maybe you can think of a function that can be called to really exit the whole program correctly?
The problem is that if you use the function `std::exit()` the program don't really exit normally - nana complain and throw, and even shows a msgbox that will block travis. The msgbox is very good for normal interactive use  but not for travis. (it is very bad to see that the OS visualice that the program crash and you need to inform about that by sending an email to Microsoft! It is much better a more or less nice msgbox from nana). It can be fixed if we find and fix why std::exit don't `std::exit()` normally.
That is why I introduced a version `API::exit_all()`. It works, closing more forms and is good for many examples, but it is also a workaround that I invented and I think it it's not very good (but is better than the main form pointer).

The whole point is that I want a method to test any program without modification, just by recompiling with a different option. For example, I want to be sure that the examples are keep always actualized. Many users and some critics try to compile old examples and then have problems.  The source text of the examples are included automatically in the documentation with many links to the classes and functions. And I don't want to explain to an absolute beginner what some artifacts for testing are. I don't want to modify the HelloWorld program to include testing, but I want to be always sure that HelloWorld compile and run without problems.
Yes, now is not very good. If you see a way to do all of that, than I agree that it is best not to change exec. I will keep thinking about, but for now that was the way I finded.

Yes - to isolate all testing functions in test.cpp is a good idea. Testing if for "advanced use" and need to be optional and separate, not a mess. But I don't know how to change one switch (a #define) and trigger a new compilation that include the test for a program that don't include it originally.

cnjinhao:
Can we replace wvl.cpp with a new defined test.cpp in travis? The test.cpp is a copy of wvl.cpp, but it has a different definition of exec(). We can move the test-related functions into test.cpp to eliminate the marco NANA_AUTOMATIC_GUI_TESTING and NANA_ADD_DEF_AUTOMATIC_GUI_TESTING. And the wvl.cpp will not be compiled in travis.

A possible definition of exec() in test.cpp is also very simple.
`
void exec()
{
	using namespace nana;
	
	timer tmr; //<nana/gui/timer.hpp>

	unsigned seconds = 10; //exit exec() after 10 seconds
	tmr.elapse([seconds]() mutable
	{
		if (0 == seconds--)
			API::exit(); //same thread of exec().
	});

	tmr.start();
	detail::bedrock::instance().pump_event(nullptr, false);
}
`
The test.cpp will not present in release distribution, I think a new complete testing feature is another issue. What do you think about this way?

I will merge the PR first.

qPCR4vir:
Sounds good, I will try.
Thank

cnjinhao:
It seemed that I found another issue, automatically testing demos without modifying demos' source code means the behavior of Nana is changed. Maybe the automatic testing is only used for internal use.

PR #123, opened 2016-03-28
FIX a drawing defect and use Geometric scheme

qPCR4vir:
Fix a drawing defect
image  
image

Also, note some reversible changes in the geometry of the list scheme (header 20 not 24, etc.)

### DIFF
--- a/include/nana/gui/widgets/listbox.hpp
+++ b/include/nana/gui/widgets/listbox.hpp
@@ -529,8 +529,18 @@ namespace nana
 				color_proxy header_floated{ static_cast<color_rgb>(0xBABBBC)};
 				color_proxy item_selected{ static_cast<color_rgb>(0xD5EFFC) };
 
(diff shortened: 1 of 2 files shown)


qPCR4vir:
I added this in the constructor of my form and got a costumized listbox.
```cpp
         SeqExpl::SeqExpl              (ThDyNanaForm& tdForm)
    {
		auto& sch = _list.scheme();
		sch.header_height = 20;
		sch.ext_w         = 2;
		sch.item_height_ex= 1;  ///< Set !=0 !!!!  def=6. item_height = text_height + item_height_ex
		sch.item_height   = sch.text_height + sch.item_height_ex;
		sch.header_mouse_spliter_area_before = 4;
		sch.header_mouse_spliter_area_after = 4 ; 

```

qPCR4vir:
Hi,
I'm using this ... is OK.
But I wanted to do the same with the `treebox`, but there you have already some type of geometrical scheme.. I need to understand that first..

PR #124, opened 2016-04-19
3 instead of 1 step by list wheel scroll

qPCR4vir:
A very simple but effective improvement

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ matrix:
             - llvm-toolchain-precise 
 
 before_install:
-  - git clone --depth=1 --branch=dev_nana_in_examples https://github.com/qPCR4vir/nana-demo.git nana-demo
+  - git clone --depth=1 --branch=hotfix-1.3 https://github.com/qPCR4vir/nana-demo.git nana-demo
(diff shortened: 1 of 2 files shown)


nabijaczleweli:
This should actually be read from OS settings, not hardcoded

cnjinhao:
Scheme is better, how about the default scheme value to be initialized with OS setting? as mentioned by @nabijaczleweli

qPCR4vir:
Hi,
yes... 

>  the default scheme value to be initialized with OS setting?

is perfect... only I dont know how.. I hope you know :-) 
I think is is best to move as much as possible to the scheme. 

It open the possibilities to adapt all (different forms factors, etc. ... for the future)

Pr0curo:
for windows you can use SystemParametersInfo with SPI_GETWHEELSCROLLLINES

i dont know about the linux/mac way :/

cnjinhao:
@qPCR4vir I will merge the changes.
@Pr0curo Thanks for the informations, I will hardcode for linux/mac.

nabijaczleweli:
It's probably acessible via some API for Mac as well

PR #130, opened 2016-05-10
textbox and combox behaviours

besh81:
textbox: when uneditable the caret is hide

combox: when uneditable (default)
- inherit the textbox behaviour
- open the lister clicking the textbox area also


cnjinhao:
Hi, the rename of arg_category has already been modified in develop branch. Normally, please commit a patch if it contains changes of some names to the develop branch. Because hotfix branch shouldn't contain any breaking change.

IMO, the caret should be still shown when the textbox is uneditable. Users can select and copy text from the uneditable textbox.

besh81:
I didn't change the arg_category, is it possible I fork a wrong repo?
Sorry, I'm new to GitHub mechanisms.

About my proposal, do you think is it possible to choose the behaviour with a define in config?
I think a lot of users will aprreciate (at least) the unselectable combox.

Ciao

cnjinhao:
Yes, a proper way is to add a flag which indicates whether to show a caret when the text_editor is uneditable. By default, textbox initializes the flag to true, combox initializes the flag to false. Additional, textbox and combox should have a method to modify the flag, so that the behavior can be changed to meet different needs.

PR #132, opened 2016-05-13
Listbox improvement.

dankan1890:
Set to 3 by default the amount of scrolling, for Windows is taken directly from the system settings.

### DIFF
--- a/source/gui/widgets/listbox.cpp
+++ b/source/gui/widgets/listbox.cpp
@@ -2287,11 +2287,16 @@ namespace nana
 					if(scroll.v.empty() || !scroll.v.scrollable(upwards))
 						return false;
 
+					int ucNumLines = 3; // default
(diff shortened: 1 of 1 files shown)


source/gui/widgets/listbox.cpp
```diff
@@ -2287,11 +2287,16 @@ namespace nana
 					if(scroll.v.empty() || !scroll.v.scrollable(upwards))
 						return false;
 
+					int ucNumLines = 3; // default
+#ifdef NANA_WINDOWS
+#include <windows.h>
```
nabijaczleweli: `#define WIN32_LEAN_AND_MEAN`

dankan1890:
I didn't add that directive for compliance with the existing code, but if cnjinhao deemed appropriate, I can add :)

PR #138, opened 2016-06-25
Dev nana fs

qPCR4vir:
more consistent use of the versions of the filesystem to make the linker happy. 
It also prepare us for the std::filesystem (no experimental).

I will need to separate the cmake  of the demos...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ matrix:
             - llvm-toolchain-precise 
 
 before_install:
-  - git clone --depth=1 --branch=dev_nana_in_examples https://github.com/qPCR4vir/nana-demo.git nana-demo
+  - git clone --depth=1 --branch=dev_nana_fs https://github.com/qPCR4vir/nana-demo.git nana-demo
   - export PATH="$HOME/bin:$PATH"
   - mkdir ~/bin
   - wget --no-check-certificate --no-clobber -O /tmp/tools/cmake https://cmake.org/files/v3.4/cmake-3.4.0-rc3-Linux-x86_64.sh || true
--- a/include/nana/c++defines.hpp
+++ b/include/nana/c++defines.hpp
@@ -59,6 +59,7 @@
 
 #		define constexpr const	//no support of constexpr until Visual C++ 2015 ? const ??
 #	endif
+#	undef STD_FILESYSTEM_NOT_SUPPORTED
 #elif defined(__GNUC__)
 #	if (__GNUC__ == 4 && __GNUC_MINOR__ < 6)
 #		define noexcept		//no support of noexcept until GCC 4.6
--- a/include/nana/filesystem/filesystem.hpp
+++ b/include/nana/filesystem/filesystem.hpp
@@ -30,6 +30,9 @@
 #ifndef NANA_FILESYSTEM_HPP
 #define NANA_FILESYSTEM_HPP
 
+//#undef  NANA_USING_NANA_FILESYSTEM 
+#if NANA_USING_NANA_FILESYSTEM 
+
 #include <nana/push_ignore_diagnostic>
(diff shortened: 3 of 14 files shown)


.travis.yml
```diff
@@ -56,7 +56,7 @@ matrix:
             - llvm-toolchain-precise 
 
 before_install:
-  - git clone --depth=1 --branch=dev_nana_in_examples https://github.com/qPCR4vir/nana-demo.git nana-demo
+  - git clone --depth=1 --branch=dev_nana_fs https://github.com/qPCR4vir/nana-demo.git nana-demo
```
cnjinhao: Does it mean that `nana/develop` branch uses different `dev_nana_*` branch of `nana-demo` in different development stages?
I found a description in your `nana-demo/dev_nana_fs` branch.
> nana/develope - nana/stable.

This is a bit weird. IMO, becuase `nana/develop` is unstable. Perhaps, the `nana-demo/stable` is more suited for `nana/master` and `nana/hotfix`.

qPCR4vir: sorry, yes, it need to be reorganized. I will think about

include/nana/c++defines.hpp
```diff
@@ -59,6 +59,7 @@
 
 #		define constexpr const	//no support of constexpr until Visual C++ 2015 ? const ??
 #	endif
+#	undef STD_FILESYSTEM_NOT_SUPPORTED
```
cnjinhao: Only undef STD_FILESYSTEM_NOT_SUPPORTED for VS2015, because VS2013 provides a <filesystem> which is not defined in namespace std::experimental.

include/nana/filesystem/filesystem.hpp
```diff
@@ -30,6 +30,9 @@
 #ifndef NANA_FILESYSTEM_HPP
 #define NANA_FILESYSTEM_HPP
 
+//#undef  NANA_USING_NANA_FILESYSTEM 
+#if NANA_USING_NANA_FILESYSTEM
```
cnjinhao: With the marco NANA_USING_NANA_FILESYSTEM , I can't use nana::experimental::filesystem::path just including <nana/filesystem/filesystem.hpp>.

qPCR4vir: Hi, Jinhao..
I suspect.. not completely sure, that the linkers have some problems, maybe bugs, that when one  compiler-unit (a cpp file with all the includes) compile with one set of functions in one namespace, and other compiler-unit, have other implementation of the same functions in another namespace it have many obscure error. 
So I chose only one implementation for all the build process. It will prefer std, them boost, and then nana. 
If you need test nana, you define `NANA_FILESYSTEM_FORCE ` in the IDE, _cmake_, _make_ or in `config.h` and it will be used... well that is the idea.
If you don't care what of the three implementation to use, you can include 

`#include <nana/filesystem/filesystem_ext.hpp>  `

that is my recommendation, because it include some functions that you use, and which are not `std `now. The non `std `features are in `nana::filesystem_ext`. all the std features are in `std::experimental::filesystem`, independently of the implementation you are using.

include/nana/c++defines.hpp
qPCR4vir: oh.. yes, it is `std::tr2::sys::path`.
Well, better we use nana.

include/nana/filesystem/filesystem.hpp
qPCR4vir: You can also add `NANA_FILESYSTEM_FORCE `in travis

cnjinhao:
Hi, I removed filesystem_selector.hpp and move its all content to the filesystem.hpp, see 1845b41019cd73ac2c0b1720f75548e235d87e01.

PR #139, opened 2016-06-25
reorganize tests

qPCR4vir:


.travis.yml
```diff
@@ -86,46 +86,46 @@ script:
   - ./animate-bmp
   - ./audio_player
   - ./background-effects
-  - ./calculator 
+  #- ./calculator   # https://travis-ci.org/qPCR4vir/nana/jobs/140238537#L1159
```
cnjinhao: > 0.12s$ ./calculator
X Error of failed request:  BadAtom (invalid Atom parameter)
  Major opcode of failed request:  18 (X_ChangeProperty)

It's strange. I tried to rebuild [travis #181](https://travis-ci.org/cnjinhao/nana/builds/140254800) many times, no X error occurs.

cnjinhao:
It looks like the XInternAtom fails to create the atom.

.travis.yml
qPCR4vir: I have no idea what that means. I can't not debug that, sorry. I have simple reported the problem... maybe you understand what is the problem? 
Here  travis #181 I have excluded all the 20 test that don't pass. The other 35 pass. The idea is that we will see if new changes break what is working now. But the link to the failure are there just in case you or somebody have the time to check what is going on. Maybe the (virtual) X server here don't work very well?

PR #146, opened 2016-07-13
c++ defines patch

dankan1890:
-  Recent versions of clang have support for make_unique.
- Check if macro "__cpp_lib_experimental_filesystem" is defined and match with the standard value.


cnjinhao:
Thanks for your PR.

PR #148, opened 2016-07-19
listbox stable_sort

qPCR4vir:
Oh !!
sorry. I'm late for 30 min ...
But... this is a very simple and small change with a very good effect.
I click to sort a column, and then other and then other and an the end all is correctly sorted.


cnjinhao:
I merged it into develop branch

qPCR4vir:
Hi, thank... is OK.
But maybe you can merge it into the hotfix too?
This was a big problem to me: when I sort a column the other columns got unsorted. Now like in Excel, you sort one, and then the other but the first is still also ordered...
It also look very "unprofessional". I jus didn't know that that was soo easy to fix.

PR #149, opened 2016-08-01
Fixed behavior of disabled items in the menus when scrolling with arrow keys.

dankan1890:
Used same behavior for the splitter.


cnjinhao:
Thanks!

PR #156, opened 2016-09-29
place::modify() now correctly updates the text to place::div_text.

dankan1890:
Note: I reverted the previous commit because it had an incorrect formatting.

### DIFF
--- a/include/nana/gui/place.hpp
+++ b/include/nana/gui/place.hpp
@@ -118,10 +118,10 @@ namespace nana
 
 		field_reference field(const char* name);///< Returns a field with the specified name.
 
-		void field_visible(const char* field_name, bool visible) const; ///<<Shows/Hides an existing field.
+		void field_visible(const char* field_name, bool visible); ///<<Shows/Hides an existing field.
 		bool field_visible(const char* field_name) const;	///<Determines whether the specified field is visible.
 
-		void field_display(const char* field_name, bool display) const; ///<Displays/Discards an existing field.
+		void field_display(const char* field_name, bool display); ///<Displays/Discards an existing field.
 		bool field_display(const char* field_name) const;	///<Determines whether the specified field is displayed.
 
 		void collocate();                     ///< Layouts the widgets.
--- a/include/nana/gui/place.hpp
+++ b/include/nana/gui/place.hpp
@@ -118,10 +118,10 @@ namespace nana
 
 		field_reference field(const char* name);///< Returns a field with the specified name.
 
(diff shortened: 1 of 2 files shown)


cnjinhao:
Thanks a lot for the patch. I have a question, why was `const` added to the setter field_display/field_visible?

dankan1890:
Because, none of the two functions change class members, and so can also be called by const class objects.

cnjinhao:
The question is
```c++
void foo(const place&); //it may implicitly modify a field's visibility

auto a = placeobj.div();
foo(placeobj);
auto b = placeobj.div();

if(a != b)
{
    //Should this condition be allowed?
}
```

dankan1890:
Test code (with 2 bug reports):
```cpp
#include <nana/gui.hpp>
#include <nana/gui/widgets/label.hpp>
#include <nana/gui/widgets/button.hpp>
#include <nana/gui/widgets/listbox.hpp>
#include <nana/gui/place.hpp>
using namespace nana;

void foo(const place &obj)
{
	obj.field_visible("lbl", false);
}

int main()
{
	form fm{ API::make_center(400, 400) };
	button btn{ fm, "Select" }, inbtn{ fm, "Make label invisible"};
	label lbl{ fm, "Label is visible" }, lbldiv{ fm };
	listbox lb{ fm };
	place placeobj{ fm };

	lb.enable_single(true, false);
	auto cat = lb.append_header("TEST");
	
	for (auto x = 0; x < 100; x++)
		lb.at(cat).push_back(std::to_string(x));
	
	lb.at(cat).at(0).select(true);

	btn.events().click([&] { 
		lb.at(cat).at(50).select(true); // Note: The listbox doesn't scroll to view the selection
	});

	inbtn.events().click([&] {
		foo(placeobj);
		lbldiv.caption(placeobj.div());
	});

	placeobj.div("<vert <mydiv><lbl><inbtn>>|<vert <lb><btn>>");
	placeobj["lb"] << lb;
	placeobj["btn"] << btn;
	placeobj["inbtn"] << inbtn;
	placeobj["lbl"] << lbl;
	placeobj["mydiv"] << lbldiv; // Note: If set the field name to "lbldiv" causes a issue in the identification of "lbl" (find_idstr function)
	placeobj.collocate();
	
	lbldiv.caption(placeobj.div());

	fm.show();
	exec();
}
```

cnjinhao:
Thanks for fixing this issue. I will make another fix for the select issue.

cnjinhao:
BTW, it would be better that if remove the `const` for the setters field_display and field_visible.

dankan1890:
Ok, no problem :)

PR #159, opened 2016-10-17
vs"15" fix

mfojtak:
nana is not compilable for Visual Studio 15.
This fixes it in c++defines.hpp.
This was not working because the _MSC_VER for VS15 exposes to 1910.

### DIFF
--- a/include/nana/c++defines.hpp
+++ b/include/nana/c++defines.hpp
@@ -102,7 +102,7 @@
 	#define _CRT_SECURE_NO_DEPRECATE
 	#pragma warning(disable : 4996)
 
-	#if (_MSC_VER == 1900)
+	#if (_MSC_VER >= 1900)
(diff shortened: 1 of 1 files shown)


qPCR4vir:
Oh!
Are you telling that the new version 1910 of the VC++ compiler which now is used in VS15 have the same error that 1900 (a known issue, tracked by an active bug)? And we need to expect that future version will always have that error?

dankan1890:
I think it is the same bug reported here  [https://connect.microsoft.com/VisualStudio/feedback/details/1403302/unresolved-external-when-using-codecvt-utf8](url)
(From second last comment looks like it's been fixed)

mfojtak:
yes, the bug is still not fixed in VS 15

cnjinhao:
Thanks a lot for your contribution, this PR has been merged into hotfix branch.

PR #164, opened 2016-11-02
fixed ICO files support under linux.

dankan1890:


dankan1890:
@toborrobot 
and with this fix you have compilation problems ?
Without this, you can see the ICO file in a picture widget (or in a listbox) under linux ?

PR #165, opened 2016-11-03
Include nana/deploy.hpp to declare to_utf8

toborrobot:
Was preventing compilation of nana library since nana::to_utf8 was undefined


cnjinhao:
Thanks for your PR, I've merged the patch to the hotfix branch.

PR #173, opened 2016-11-28
fixed 'invisible' checks and images in listbox with cells with a custom format enabled.

dankan1890:


dankan1890:
I apologize for the first commit, I mixed the code with some changes that I'm still testing.
Now it should work :)

cnjinhao:
Hi, the scheme->item_bordered and scheme->item_border are undefined, it seems that listbox.hpp is missing.

cnjinhao:
Hi, the scheme->item_bordered and scheme->item_border are undefined, it seems that listbox.hpp is missing.

PR #181, opened 2017-01-19
Use target_include_directories to propagate includes

redagito:
When using nana as a submodule, this will automatically add the include directories from nana to all targets depending on nana.
Basically this removes the need to add something like
`
target_include_directories(${PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/external/nana/include)`
to the CMakeLists file of a project using nana


cnjinhao:
Hi, thanks for your PR.

PR #186, opened 2017-02-18
Fix non-unicode char (and typo)

alexkaratarakis:
`nana\src\nana\include\nana\gui\programming_interface.hpp` line 291 has a non-unicode apostrophe. This creates a large number of warnings (also there is typo).

Related: https://github.com/Microsoft/vcpkg/issues/688


alexkaratarakis:
GIthub's preview doesn't show the non-unicode char.
It is in line 291 at the diamond <>:
`Refreshes the specified window and all it<>s children windows`

cnjinhao:
Thanks a lot for your PR!

PR #206, opened 2017-05-14
CMake: fix NANA_LINKS for MSVC

codicodi:
This turns `NANA_LINKS` variable to a list, thus allowing CMake/MSVC build with dependencies (png, jpeg) enabled.
MSVC does not really understand `-l<name>` flags and concatenating `${<NAME>_LIBRARIES}` into a giant string breaks CMake's metadata passed in that variables.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,7 +100,7 @@ endif(WIN32)
 if(APPLE)
     add_definitions(-DAPPLE)
     include_directories(/opt/X11/include/)
-    set(NANA_LINKS "${NANA_LINKS} -L/opt/X11/lib/ -liconv")
+    list(APPEND NANA_LINKS -L/opt/X11/lib/ -liconv)
     set(ENABLE_AUDIO OFF)
 elseif(UNIX)
     add_definitions(-Dlinux)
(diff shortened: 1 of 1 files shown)


qPCR4vir:
Hola,
I can not remember, and to bad from me, I did not commented... this seems to revert https://github.com/cnjinhao/nana/commit/972ecc97ac030e917d4309f4ffe0f52609687d96
 but it seems to work for you with VS and it is working in Travis/linux. 

It could be a problem in cmake/gcc/windows/CLion? Some times, in some situations it was an error. Something with "uninitialized" NANA_LINKS that produced an error with append? or introduction of ";" in some command?

codicodi:
I've tried several build option combinations and several compilers and it seems to works fine.

cnjinhao:
Hi, guys.
I don't have much experience on CMake, but I know that, in some situation, a very simple stuff that seemsright, but it is probably wrong. Nana is required to works on various compilers and environments, it makes the CMakefile a bit complicated. But I have not enough experience to judge whether the change is correct(Maybe it is wrong in a old version CMake, in a certain system). I hope someone can give a comprehensive explaination.

Thanks in advance.

loaden:
I can't explain more because my English issue. But this patch sounds good to me.

cnjinhao:
I merged this PR. If there is any problem, please let me know.

PR #208, opened 2017-05-20
Fix VS2017 project win32 config

FirstTimeInForever:
For some reason win32 configuration was broken. There were no include paths and other stuff were just set to their defaults.


cnjinhao:
Thank you for your PR!

PR #221, opened 2017-06-25
Fixed build error

Red-Portal:
compiler didn't understand std::function
included ```<functional>``` as a fix


cnjinhao:

PR #222, opened 2017-06-26
Fixed stdc++.hpp build error on gcc 7

Red-Portal:
included algorithm in c++defines.hpp
this fixes std::clamp's feature test which was not working.

__cpp_lib_clamp is defined inside ```<algorithm>```.
Not include ```<algorithm>``` before the feature test introduced the problem.


cnjinhao:
Thanks for you patch!

PR #224, opened 2017-07-06
Fix codeblocks build project

JamesBremner:
So that it will build 'out of the box'.  Set c++11.  Define
enable_std_make_unique.  Define various _WIN32*   Add boost library to
search folders.


cnjinhao:
Thank you for your PR!

Issue #226, opened 2017-07-12
nana::textbox should handle mouse double-click


os12:
Double-click should select the word under cursor (at least on Windows). This started in #225 and comes down to the following modes:
* read-write
* read-only with `enable_caret()`

PR #227, opened 2017-07-12
Added Windows-style mouse double-click to the textbox

os12:
- the word under cursor is selected with double-click now
 - tested in both editable and read-only modes

This addresses #226.

### DIFF
--- a/source/gui/widgets/skeletons/text_editor.cpp
+++ b/source/gui/widgets/skeletons/text_editor.cpp
@@ -440,7 +440,7 @@ namespace nana{	namespace widgets
 						break;
 				}
 
-				return *colored_areas_.emplace(i, 
+				return *colored_areas_.emplace(i,
 						std::make_shared<colored_area_type>(colored_area_type{line_pos, 1, color{}, color{}})
 					);
 			}
@@ -1129,7 +1129,7 @@ namespace nana{	namespace widgets
 				}
 			}
 		}
-		
+
 		colored_area_access_interface& text_editor::colored_area()
 		{
 			return impl_->colored_area;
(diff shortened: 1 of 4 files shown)


os12:
Or should this go into origin/develop?

cnjinhao:
Thanks for your PR, this improvement doesn't impact on the hotfix, it is OK for both 'develop' and 'hotfix'.

source/gui/widgets/skeletons/text_editor.cpp
```diff
@@ -1456,6 +1456,10 @@ namespace nana{	namespace widgets
 					select_.ignore_press = (arg_focus::reason::mouse_press == arg.focus_reason);
 				}
 			}
+
+			// Reset the selection as it is sticky in the "double click" path.
+			if (!arg.getting)
+				select(false);
```
cnjinhao: Is it possible that deselect only if the text is selected by double-click?

cnjinhao: Maybe it should keep the selection even though it loses focus.

os12: That's what the first commit did. However, it is then possible to select a word from several `textbox` controls on a single form. That looks a bit weird (and that not the way things work when you select text by dragging the mouse).

cnjinhao: Very nice tip! Let's keep the selection even though it loses focus. The issue where multiple textboxes on a single form would be marked as a known issue and will be fixed in 1.6.
The new fix will be like this. When a text is selected, deselect selection in other textbox.

os12:
OK, I've removed the second (selection-related) commit.

cnjinhao:
Thanks a lot for the contribution!

PR #234, opened 2017-07-27
Inputbox set minimum width entry field

JamesBremner:
V1.5.4 increased the default minimum width of the entry fields in inputbox to a sensible value of 100 pixels.  However, we sometimes need even wider fields  than this, but it would be weird to make the default any wider than 100 pixels.  This add a setter to increase the width when needed.

### DIFF
--- a/include/nana/gui/msgbox.hpp
+++ b/include/nana/gui/msgbox.hpp
@@ -46,7 +46,7 @@ namespace nana
 		/// Construct that creates a message box with a specified title and default button.
 		msgbox(const ::std::string&);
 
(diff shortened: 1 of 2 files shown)


JamesBremner:
Usage example:

        form fm;

        // construct text input the expects long inputs
        inputbox::text test(
            "very very very very long prompt string",
            "very long default");
            
        // construct inputbox that provides especially wide input fields
        inputbox inbox(fm,"Enter");
        inbox.min_width_entry_field( 250 );
        
        //show 'em what we have
        if (inbox.show(test) )
        {
        }

cnjinhao:
Thanks for your PR!

PR #262, opened 2017-09-25
Added a textbox::append method

wjiali6:
Added a textbox::append method which accepts std::wstring text


cnjinhao:
Thanks a lot for your PR!

Issue #277, opened 2017-11-09
Build error


KingDuckZ:
I'm just trying the "hello world" tutorial, but I'm getting a few errors. The first one is that the c++14 standard is not inherited by my cmake project when I link to nana. To fix that I removed the explicit -std options from the cmake file as well as the CXX_STANDARD property (I don't know if there's a way to make that a public property so it would get inherited) and replaced it with:

    target_compile_features(${PROJECT_NAME}
        PUBLIC cxx_nullptr
        PUBLIC cxx_range_for
        PUBLIC cxx_lambdas
        PUBLIC cxx_decltype_auto
        PUBLIC cxx_defaulted_functions
        PUBLIC cxx_deleted_functions
        PUBLIC cxx_auto_type
        PUBLIC cxx_decltype_incomplete_return_types
        PUBLIC cxx_defaulted_move_initializers
        PUBLIC cxx_noexcept
        PUBLIC cxx_rvalue_references
    )

this got me past the first problem and I'm wondering if this couldn't help with issue #269 as well?

The second problem is a linker error:

/usr/lib/gcc/x86_64-pc-linux-gnu/5.4.0/../../../../x86_64-pc-linux-gnu/bin/ld: lib/nana/libnana.a(platform_abstraction.cpp.o): undefined reference to symbol 'FcConfigAppFontAddFile'
/usr/lib64/libfontconfig.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status

which I'm looking into right now.

Mizux:
Seems to be defined by FontConfig
src: https://gitlab.freedesktop.org/fontconfig/fontconfig/blob/master/fontconfig/fontconfig.h#L448

Which version do you use ?

qPCR4vir:
I'm preparing a PR for that and more: https://github.com/qPCR4vir/nana/commit/de2ec3928da7cbe9589db36f7406cd1fda7baa67

Cold you test my new version of [CMakeLists.txt](https://github.com/qPCR4vir/nana/blob/CMake/CMakeLists.txt) ?

qPCR4vir:
hi @KingDuckZ ,
It is my bad and a shame that I didn't saw this PR and your explanation a year ago. More exactly, I probably read it but I was unprepared for it and just ignore it. Sorry. By trying to fix by myself the problems I kome back here to see that you have alrready pointed to most of what I was using and a few more.
[...]

PR #278, opened 2017-11-10
Develop

KingDuckZ:
I restyled the cmake file while fixing the problem I had in issue #277.
I suggest the use of target_include_directories (instead of include_directories) and target_compile_definitions (instead of add_definitions) as they allow more control over the properties being specified. For example client programs linking to nana will now inherit the public include paths and macro definitions, but not the private ones (cmake will still make sure all the needed pieces are there at link-time).

Some of these changes required me to bump cmake version to 3.1, which should be available pretty much everywhere at this point.

I'm also using the variables provided by the cmake scripts to link to the required libraries and they often contain absolute paths, which made NANA_LINKS look very messy. Also note that the new cmake find scripts tend to define an imported target (for example PNG::PNG, Boost::Boost) to link to, which is very handy and reduces the possibility of mistakes. This, however, would require extracting the actual name of the library from the target in order to be added to NANA_LINKS, and I'm not sure it is worth the extra code. For those reasons I decided to just get rid of the NANA_LINKS library.

As a personal preference I don't like to see too much output in my cmake, unless I want to (in which case I can compile with make VERBOSE=1), so I would be for getting rid of all the print statements at the end of the cmake file. This should remove quite a bit of clutter as well.

**Important**: I'm only able to test this on Linux, so it's possible my changes introduce problems to non-Linux platforms. Somebody with access to other OS should have a second look at my changes.

I'm open to discussions on how to further improve the cmake file.


cnjinhao:
Thanks for your PR! This PR has been referred by qPCR4vir's(#342 ).

qPCR4vir:
It is my bad and a shame that I didn't saw this PR and your explanation a year ago. More exactly, I probably read it but I was unprepared for it and just ignore it. Sorry. By trying to fix by myself the problems I kome back here to see that you have alrready pointed to most of what I was using and a few more. 
I think now left only to use the Boosr::boost thing, but a got some problem and I have read that some time Boost go to fast for cmake and one need to use the other aproach. 
I hope you will keep helping. 
Thank!

PR #281, opened 2017-11-25
fix compilation with boost::filesystem

pavelxdd:
see qPCR4vir/nana-demo#13 for related nana-demo changes, without them travis build is always going to fail on tests.

Travis build succeeded with this PR and changes to nana-demo applied:
https://travis-ci.org/pavelxdd/nana/builds/308570743

CMakeLists.txt:
---------------------------
Until now build with boost::filesystem was completely broken,
since cmake exported definitions with wrong name prefixes, and
nana always defaulted to internal filesystem implementation.

After fixing the boost definitions, a number of errors came up
due to incompatibility of boost::filesystem with nana and std
filesystems. This commit tries to fix them all.

filesystem.cpp, filesystem.hpp, filebox.cpp:
--------------------------------------------
boost::filesystem doesn't have a file_time_type, so declared
it in the filesystem.hpp header.

boost::filesystem::last_write_time has a return type std::time_t
unlike the other two implementations of this function in nana
and std, so added ifdef to convert the result to file_time_type.

fixed build on gcc-4.9, since it doesn't have a std::put_time
function, included <nana/stdc++.hpp> in that case.

boost::filesystem::file_type types have different names than
std::experimental::filesystem::file_type types, fixed it by
creating an enum class file_type with the same type names as
in std::experimental::filesystem::file_type. This fix
requires static_cast from functions results to internal file_type,
since boost file_type and std file_type a different enum classes.

changed switch to if, bacause old gcc fails on converting
enum class members to int.

stdc++.hpp:
-----------
added ifndef guards to prevent errors on multiple includes of this header.

wvl.cpp:
--------
added boost/chrono.hpp include for the cases when std::thread is not
available.

travis:
-------
added boost system, thread, chrono libs to install, they are needed
for the nana-demo to compile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,9 @@ matrix:
             - libx11-dev
             - libxft-dev
             - libboost-filesystem-dev
+            - libboost-system-dev
+            - libboost-thread-dev
+            - libboost-chrono-dev
           sources: 
             - ubuntu-toolchain-r-test
             
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,7 +238,7 @@ elseif (NANA_CMAKE_STD_FILESYSTEM_FORCE)
     add_definitions(-DSTD_FILESYSTEM_FORCE)
 elseif (NANA_CMAKE_FIND_BOOST_FILESYSTEM OR NANA_CMAKE_BOOST_FILESYSTEM_FORCE)
     if (NANA_CMAKE_BOOST_FILESYSTEM_FORCE)
-        add_definitions(-DNANA_BOOST_FILESYSTEM_FORCE)
+        add_definitions(-DBOOST_FILESYSTEM_FORCE)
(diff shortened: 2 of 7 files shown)


pavelxdd:
Updated PR.

cnjinhao:
Hi
Thanks for your contribution. It has still an error on GCC4.9
programming_interface.cpp.o): undefined reference to symbol '_ZN5boost6system15system_categoryEv'

//usr/lib/x86_64-linux-gnu/libboost_system.so.1.54.0: error adding symbols: DSO missing from command line

pavelxdd:
see qPCR4vir/nana-demo#13 for related nana-demo changes, without them travis build is always going to fail on tests.

Travis build succeeded with this PR and changes to nana-demo applied:
https://travis-ci.org/pavelxdd/nana/builds/308570743

cnjinhao:
Thanks for your explaination, I got it

PR #285, opened 2018-01-14
Added missing std::move, within tabbar insert overload for std::string. [Develop Branch]

5cript:
The diff is self-explanatory. Fixed insert overload for tabbar.

PR #286, opened 2018-01-14
Added missing std::move [Hotfix Branch]

5cript:
I also provided the same change as  #285 to the hotfix, that I requested for develop.


cnjinhao:
Thanks for your PR!

PR #288, opened 2018-01-22
FreeBSD and general POSIX

Unitrunker:
Changes:

POSIX_NANA for common code bits.
thread_t for the thread id (unsigned didn't work on 64 bit).
NANA_UNICODE defaults to "UTF-32LE" unless you say otherwise on the command-line.
Primitive audio for OSS compliant systems like FreeBSD.
nana::system::open_url works - if you have Firefox. Code is written so that generalizing to detect browsers is easy.

Risks:

I have not tested Windows or Linux. I estimate risk to Windows platform is very low. I am more concerned about Linux in case I over-generalized POSIX commonality. Mac OSX should be no worse than before. I suspect the unicode fix here also applies to Mac.

I can test on Windows 7/8/10 and Linux (either Slackware or Ubuntu) but will need more time. I cannot test Mac OSX. 

The use of posix_spawn for open_url should be reviewed for any security concerns.


Unitrunker:
The linker needs '-lboost_system' for gcc 4.9.

Unitrunker:
The linker needs '-lboost_system' for gcc 4.9.

qPCR4vir:
Hi Jinhao,
I suggest you to open a new "feature" branch based on your develop branch to hold this. We need that someone else more test this, to be included in the next version.
Great contribution.
Thank @Unitrunker

qPCR4vir:
Our user **blab** [compiled this in MacOSX](http://nanapro.org/en-us/forum/index.php?u=/topic/141/ggsupporting-maxosx)  !!

cnjinhao:
I started a new `feature-freebsd-posix` branch for this PR

Unitrunker:
awesome!​

PR #290, opened 2018-01-24
Make more Linux distros happy

buck-yeh:
Specifically, making Nana C++ in Fefora 27 will complain about missing <ft2build.h>, and so will Solus.

The following distros have been tested against the change:
Arch Linux, Manjaro, Ubuntu, Fedora, Solus


cnjinhao:
Thanks for your PR!

PR #291, opened 2018-01-26
Feature toolbar proxy

5cript:
I am sorry, git diff is drunk :(

toolbar.hpp line 36-42 changed and 99-100
toolbar.cpp line 424-427.

I added a struct that allows you to disable a toolbar item right after appending it.


5cript:
Actually such a thing could be done in several places, not just the toolbar.

cnjinhao:
Thanks for your PR!

PR #294, opened 2018-01-29
develop-c++17: Fix travis test/demos with gcc 4.9 and boost 1.54

qPCR4vir:
Or just merge your develop or hotfix back to develop-c++17.


cnjinhao:
Thanks!

Issue #295, opened 2018-01-29
Cross-compiling to Windows on Linux is broken without mingw-std-threads 


dnso86:
https://github.com/cnjinhao/nana/blob/a653a7a3fa6de698b67693bbe66e8bdeed5507f7/include/nana/c%2B%2Bdefines.hpp#L225

Hi there, this change broke my makefile for cross-compiling Nana with MinGW - perhaps we should give this a second thought? I totally want to use Boost::thread, without changing anything in this file. :)

qPCR4vir:
Hi, coming from [here](https://github.com/cnjinhao/nana/commit/252b8cbf650eea663b35d6198e1127f6d8562cd3):

> I need some help to understand what you mean.
You are in Linux using MinGW to create windows binaries, and your MinGW dont have std-threads, so you want to use something else for threads. Ok?
Let me know where I'm wrong:
`#ifdef NANA_MINGW` is true, and `STD_THREAD_NOT_SUPPORTED` will be defined.
`#  if __has_include(<mutex>)` is false and  `STD_THREAD_NOT_SUPPORTED` is defined.
With `STD_THREAD_NOT_SUPPORTED` defined nana will pick some other implementation, not the std.
IF there is a problem and you see how to fix it, tell me or make a PR and we will fix it in the repository.
Thank

Now I understand you want to force nana to use Boost::thread. 
We will need to to add something like `BOOST_FILESYSTEM_FORCE` in [filesystem](https://github.com/cnjinhao/nana/blob/master/include/nana/filesystem/filesystem.hpp#L52) and [CMakeLists](https://github.com/cnjinhao/nana/blob/master/CMakeLists.txt#L57) if using cmake, to modify [nana/std_thread](https://github.com/cnjinhao/nana/blob/master/include/nana/std_thread.hpp#L5) to force execution of [#include <boost/thread.hpp>](https://github.com/cnjinhao/nana/blob/master/include/nana/std_thread.hpp#L15)

dnso86:
Hi qPCR4vir, thanks for the answer! Yes, I want to force Nana to do that, but before that change which I pointed out, it worked correctly for cross-compilation. Let me look into this tomorrow. Also, I would be glad to do the PR for this.

dnso86:
So, I looked into my project:

- The `mutex` header is present at `/usr/lib/gcc/i686-w64-mingw32/6.3-win32/include/c++/mutex` as a part of the respective Debian package
[...]

PR #296, opened 2018-01-31
Checking whether MinGW really supports threads

dnso86:
Related to https://github.com/cnjinhao/nana/issues/295, see the detailed description in the issue.

Further investigation revealed that unless the MinGW binary is built with `--enable-threads=posix` - which is not the default - it can't implement threading as we expect in Nana, despite the `mutex` header being present.

I tested the changes with **i686-w64-mingw32-g++**, gcc version 6.3.0 20170516 (GCC), targeting Windows x86 from Linux.

### DIFF
--- a/include/nana/c++defines.hpp
+++ b/include/nana/c++defines.hpp
@@ -171,9 +171,9 @@
 	#endif
 #endif
 
-//Assume the std::thread is not implement on MinGW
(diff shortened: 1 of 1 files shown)


cnjinhao:
Thanks for your PR.

PR #302, opened 2018-02-23
Added listbox::clear_headers() and menubar::clear()

besh81:
Added listbox::clear_headers() to remove all the columns from the listbox
Added menubar::clear() to remove all the items from menubar

### DIFF
--- a/source/gui/widgets/listbox.cpp
+++ b/source/gui/widgets/listbox.cpp
@@ -5328,6 +5328,7 @@ namespace nana
 		{
 			internal_scope_guard lock;
 			auto & ess = _m_ess();
+			ess.lister.erase();
 			ess.header.clear();
 			ess.update();
 		}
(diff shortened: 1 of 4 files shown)


source/gui/widgets/listbox.cpp
```diff
@@ -5319,6 +5324,14 @@ namespace nana
 			return pos;
 		}
 
+		void listbox::clear_headers()
+		{
+			internal_scope_guard lock;
+			auto & ess = _m_ess();
+			ess.header.clear();
```
cnjinhao: I think it is better that all items are deleted as well

besh81: Yes I think it's better too :+1:

cnjinhao: So, would you like to update the PR? Im looking forward to it.

besh81:
I've added this line:
`ess.lister.erase();`
that should erase all the items and categories.

source/gui/widgets/listbox.cpp
besh81: Added ess.lister.erase(); to erase all the items.

cnjinhao:
Thanks for your PR!

PR #323, opened 2018-07-08
New textbox::getline overload returning an optional<string>

rbrugo:
I've added two new overloads for `textbox::getline`:
```cpp
std::optional<std::string> textbox::getline(std::size_t pos) const;
std::optional<std::string> textbox::getline(std::size_t line_index, std::size_t offset) const;
```
Using `std::optional` the user can have as a result either the string or a failure (the empty state) without need to create a new `std::string` to pass by reference.


cnjinhao:
Thank you for your PR!

Issue #338, opened 2018-09-23
Nana pollutes std namespace ( with its own optional).


JenioPY:
While compiling nana on Visual Studio (15.8.5) with `std:c++17` it instead of including `<optional>` defines the nana::optional and pollutes the std namespace with:
```c++
namespace std
{
	using nana::optional;
}
```
After that if I try to include `<optional>` in my project i'll get multiple declaration compile errors is not surprising considering above.
Files involved are:
`c++defines.hpp`
`optional.hpp`

cnjinhao:
Thank you for reporting and fixing this bug! A new patch has been checked into the hotfix branch.

PR #339, opened 2018-09-23
Fixed condidtion for defining _nana_std_otpional.

JenioPY:
This is my quick fix regarding #338. I believe this condition was broken but I might be wrong.


cnjinhao:
Thanks for your pull request. The bug#338 is caused by the __cplusplus which defined in Visual Studio remaining at value `199711L` if `/Zc:__cplusplus` is not given. Your patch works in Visual Studio 2017 but doesn't work in older version of VS. I will close this PR and commit a new patch. Thank you again. I'm looking forward to your other fixes.

PR #342, opened 2018-10-02
Cmake dev

qPCR4vir:
Just take a first look at this by creating a new dev branch.
The main problem was that nana-demo CMakeLists was ugly and a duplication of nana CMakeLists. I'm trying to use "modern" cmake with a target-centric approach. It is working correctly with https://nuwen.net/mingw.html which come with gcc 8.1 and with precompiled Boost.
But Im having problems with our Travis-IC: [wvl.cpp:(.text+0x3d9): undefined reference to `pthread_create'](https://travis-ci.org/qPCR4vir/nana-demo/jobs/427235069#L1323)
And with JPEG: https://github.com/qPCR4vir/nana-demo/issues/15#issuecomment-425999687

You will need to clear (only one time) the cache in travis to get the new version of cmake really installed. 

I saw now the old PR https://github.com/cnjinhao/nana/pull/278. I will try to use it to solve some of my problems.


qPCR4vir:
Run OK now

cnjinhao:
I got 2 errors that 2 examples `example.button` and `stretch_image` failed to open BMP file(../Examples/bground.6states.bmp). But, I ran these 2 examples in my local linux system, no error occurred.

PR #346, opened 2018-10-09
optional verbose nana cmake

qPCR4vir:
Recomended by @KingDuckZ in https://github.com/cnjinhao/nana/pull/278
Silent warnings from boost


qPCR4vir:
The API file filebox.hpp already depends on filesystem for folderbox API: 
then lets filebox API take a fs::path too for init_path and init_file?? It can make it easier to check

source/gui/filebox.cpp
```diff
@@ -1278,17 +1290,22 @@ namespace nana
...
-		brw.ulFlags = BIF_RETURNONLYFSDIRS | BIF_NEWDIALOGSTYLE;
-		brw.lpfn = browse_folder_callback;
-
+		auto title = to_wstring( impl_->title ) ;
 		std::wstring init_path = impl_->init_path.wstring();
-		brw.lParam = reinterpret_cast<LPARAM>(init_path.c_str());
-		auto pidl = ::SHBrowseForFolder(&brw);
 
+		// https://docs.microsoft.com/en-us/windows/desktop/api/shlobj_core/ns-shlobj_core-_browseinfoa
+        BROWSEINFO brw       = { 0 };
```
beru: Here, 8 spaces are used for indentation. Using 2 tabs might be good to keep consistency.

include/nana/gui/filebox.hpp
```diff
@@ -103,6 +103,11 @@ namespace nana
 		{
 			return show();
 		}
+
+		/// Set a new title for the dialog
+		/// @param	string	a text for title
+		/// @return the old title.
+		::std::string title( ::std::string new_title);
```
beru: The argument type should be const reference. I hope `std::string_view` will be used in future though.

include/nana/gui/filebox.hpp
```diff
@@ -9,7 +9,7 @@
  *
  *	@file nana/gui/filebox.hpp
  *  @author Jinhao
- *  @brief a dialog to chose file(s), implemented "native" in windows but using nana for X11 
+ *  @brief dialogs to chose file(s) or a directory, implemented "native" in windows but using nana for X11
```
beru: How about splitting files for filebox and folderbox?

beru: `choose` should be used here instead of `chose`.

cnjinhao: OK

include/nana/gui/filebox.hpp
cnjinhao: OK

cnjinhao:
Thanks for your PR!

Issue #341, opened 2018-09-27
Multiple file selection in nana::filebox


quasinormalized:
Am I merely overlooking this feature/option?
I often utilize a file dialog's ability to open multiple (but not necessarily all) files from a given folder (e.g. using the Shift/Ctrl keys in MSWin).

#### Feature request:
`nana::filebox` with option to select single file or multiple files (selected files or entire folder) when `is_open_mode == true`. I believe the implementation in MSW is: _[IFileOpenDialog](https://docs.microsoft.com/en-us/windows/desktop/api/shobjidl_core/nn-shobjidl_core-ifileopendialog)_.

Perhaps a member function (::file**s**()?) could return a vector of file paths?

![Open multiple files in VS](https://user-images.githubusercontent.com/43644768/46160540-3a809180-c240-11e8-8f39-7b2c85dcd32d.png)

beru:
Talking about nana's implementation for Windows,
`nana::filebox::show()` does not use COM based `IFileOpenDialog` yet.
`nana::folderbox::show()` uses it though.

`nana::filebox::show()` still uses outdated `GetOpenFileName`.

https://github.com/cnjinhao/nana/blob/dbd8a4a6910aa3b81b77989a2d1866fdb26e4f6f/source/gui/filebox.cpp#L1165

quasinormalized:
Yeah, I ended up with this for testing. It uses ATL's smart-pointers & seems to work fine, but the MinGW & OSX snippets that Nana would need are beyond me.

Header:

    #include <exception>
    #include <vector>
    #include <string>
    
    namespace ms
    {
        struct UtilizeCOM
        {
            UtilizeCOM()
            {
                // From MSDN: "it is a good idea to set the COINIT_DISABLE_OLE1DDE flag in the dwCoInit
                //  parameter. Setting this flag avoids some overhead associated with Object Linking and
                //  Embedding (OLE) 1.0, an obsolete technology"
                if (FAILED(::CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE)))
                    throw std::runtime_error("Couldn't initialize Windows' (OLE-less) COM");
            }
            ~UtilizeCOM() { ::CoUninitialize(); }
            UtilizeCOM(const UtilizeCOM&)			= delete;
            UtilizeCOM& operator=(const UtilizeCOM&)	= delete;
        };

[...]

PR #349, opened 2018-10-28
Allow multiple file selection with nana::filebox on Windows platform

beru:
Hi, this PR adds the feature mentioned in #341.

There can be many decisions when implementing multiple file selection feature such as,
- how to hold multiple filenames internally.
- should `files` method return only filenames? or fullpaths?

I only added support for Windows platform so it makes this feature kinda half-baked...

### DIFF
--- a/include/nana/gui/filebox.hpp
+++ b/include/nana/gui/filebox.hpp
@@ -68,8 +68,12 @@ namespace nana
         };
 
 
-		::std::string path() const;
-		::std::string file() const;
+		const ::std::string& path() const;
+		const ::std::string& file() const;
+#if defined(NANA_WINDOWS)
+		const ::std::vector<::std::string>& files() const;
+		void allow_multi_select(bool allow);
+#endif
 
 		/// Display the filebox dialog
 		bool show() const;
--- a/source/gui/filebox.cpp
+++ b/source/gui/filebox.cpp
@@ -995,7 +995,12 @@ namespace nana
 		window owner;
 		bool open_or_save;
 
+#if defined(NANA_WINDOWS)
+		bool allow_multi_select;
+		std::vector<std::string> files;
+#else
 		std::string file;
+#endif
 		std::string title;
 		std::string path;
 		std::vector<filter> filters;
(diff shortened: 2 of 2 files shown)


beru:
There's an error with Travis-CI build.
https://travis-ci.org/cnjinhao/nana/builds/447377153#L1254

> CMake Error at CMakeLists.txt:18 (cmake_minimum_required):
  CMake 3.12 or higher is required.  You are running version 3.4.0-rc3

https://github.com/qPCR4vir/nana-demo/blob/develop/CMakeLists.txt#L18
https://github.com/qPCR4vir/nana-demo/blob/794e471ede00e92316be1b49d4f08dc37258c3a0/CMakeLists.txt#L18
> cmake_minimum_required(VERSION 3.12)

To resolve the error, this line needs to be updated.
https://github.com/cnjinhao/nana/blob/852aa698ed65d03733d03eddfbdb8fe206bcac3a/.travis.yml#L47
```
 - wget --no-check-certificate --no-clobber -O /tmp/tools/cmake https://cmake.org/files/v3.4/cmake-3.4.0-rc3-Linux-x86_64.sh || true
```

beru:
This is the test code I used to check the feature.

```cpp
#include <nana/gui.hpp>
#include <nana/gui/filebox.hpp>

int main()
{
    using namespace nana;
    filebox fb(true);
    fb.add_filter("Text File", "*.txt;*.text;*.doc");
    fb.add_filter("All Files", "*.*");
    fb.allow_multi_select(true);
    
    if(fb())    //When user clicked 'OK'
    {
        auto file = fb.file();
        auto& files = fb.files();
        auto path = fb.path();

        printf("path : %s\n", path.c_str());
        for (auto& f : files)
        {
            printf("file : %s\n", f.c_str());
        }
    }
    else
    {
        auto file = fb.file();
        printf("file : %s\n", file.c_str());
    }
}
```

beru:
It looks like `cnjinhao:develop` branch is outdated, I should target `develop-1.7` branch...

PR #352, opened 2018-11-07
FIX crash while deleting a selected tree node

qPCR4vir:
replace https://github.com/cnjinhao/nana/pull/350
see picture there


cnjinhao:
Thank you for your PR!

PR #357, opened 2018-11-13
Resort column range

qPCR4vir:
I need to reorder a range of columns based on the values in some row. 
This is a possible implementation for that "feature".
It is working, but,  of course review it carefully.

### DIFF
--- a/include/nana/gui/widgets/listbox.hpp
+++ b/include/nana/gui/widgets/listbox.hpp
@@ -1461,6 +1461,9 @@ the nana::detail::basic_window member pointer scheme
 		/// Returns the number of columns
 		size_type column_size() const;
 
+		/// Move column to view_position
+		void move_column(size_type abs_pos, size_type view_pos);
+
 		void column_resizable(bool resizable);
 		bool column_resizable() const;
 		void column_movable(bool);
@@ -1517,7 +1520,8 @@ the nana::detail::basic_window member pointer scheme
 		
 		///Sets a strict weak ordering comparer for a column
 		void set_sort_compare(	size_type col,
-								std::function<bool(const std::string&, nana::any*, const std::string&, nana::any*, bool reverse)> strick_ordering);
+								std::function<bool(const std::string&, nana::any*,
+								                   const std::string&, nana::any*, bool reverse)> strick_ordering);
 
 		/// sort() and ivalidate any existing reference from display position to absolute item, that is: after sort() display offset point to different items
 		void sort_col(size_type col, bool reverse = false);
--- a/include/nana/gui/widgets/listbox.hpp
+++ b/include/nana/gui/widgets/listbox.hpp
@@ -1464,7 +1464,16 @@ the nana::detail::basic_window member pointer scheme
 		/// Move column to view_position
 		void move_column(size_type abs_pos, size_type view_pos);
 
-		void column_resizable(bool resizable);
+        /// Sort columns in range first_col to last_col inclusive using the values from a row
+        void reorder_columns(size_type first_col,
+							 size_type last_col,
+							 index_pair row, bool reverse,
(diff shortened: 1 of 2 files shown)


include/nana/gui/widgets/listbox.hpp
```diff
@@ -1461,7 +1461,19 @@ the nana::detail::basic_window member pointer scheme
 		/// Returns the number of columns
 		size_type column_size() const;
 
-		void column_resizable(bool resizable);
+		/// Move column to view_position
+		void move_column(size_type abs_pos, size_type view_pos);
+
+        /// Sort columns in range first_col to last_col inclusive using the values from a row
+        void reorder_columns(size_type first_col,
```
beru: Whitespace letters are used for indentation here, how about considering tabs?

source/gui/widgets/listbox.cpp
```diff
@@ -130,17 +130,18 @@ namespace nana
 				struct column
 					: public column_interface
 				{
-					native_string_type caption;
-					unsigned width_px;
-					std::pair<unsigned, unsigned> range_width_px;
+					native_string_type caption;                     //< header title
+					unsigned width_px;                              //< column width in pixels
+					std::pair<unsigned, unsigned> range_width_px;   //< allowed witdh
```
beru: Found a typo here. `witdh` should be `width`

source/gui/widgets/listbox.cpp
```diff
@@ -542,6 +543,24 @@ namespace nana
...
+							               [&](const column& c){return col==c.index;});
+
+					if (i==cont_.end()) return;
+
+					auto col_from = *i;
+					cont_.erase(i);
+					cont_.insert(cont_.begin()+ view, col_from);
+
+				}
 				/// move the col originaly at "from" to the position currently in front (or after) the col originaly at index "to" invalidating some current index
```
beru: Found a typo here. `originaly` should be `originally`

include/nana/gui/widgets/listbox.hpp
qPCR4vir: Hi thanks! I could agree, but for format style questions, I assume Jinhao's decisions. At least until we make a general style guide. On the other hand see what GitHub does with tab indentation in the following lines

qPCR4vir:
> @qPCR4vir Can you provide sample code to check the feature? That'd be very helpful.

I understand that what you want is an small but full demo... I will try that, but by now this the real use:

https://github.com/qPCR4vir/ThDySec/blob/master/include/ThDy_DNAHybrid.Nana/TableResults.h#L252

The click event of the button _byCol ( <-- ) reorder the columns from 1 to last (exclude 0 column with the names) with have homogenic numeric comparable values, based on the row currently selected by the user. It shows the "best" results near the names at the left of a very large table. No more need to scan the table for the best values for that row

        _byCol.events().click([&]()
          {
              auto sel = _list.selected();
              if (sel.size() != 1 ) return;
              _list.reorder_columns(1,                      // size_type first_col
                                    _table->totalCol(),    // size_type last_col
                                    sel[0], true,          // index_pair row, bool reverse ,
                                    [&](const std::string &cell1, List::size_type col1,
                                        const std::string &cell2, List::size_type col2,
                                        const nana::any *rowval,
                                        bool reverse) -> bool {
                                        const Index *I = nana::any_cast<Index>(rowval);
                                        auto &v = *I->table->val;
                                        bool r = (v.val(I->row, col1 - 1) <= v.val(I->row, col2 - 1));
                                        return reverse ? !r : r;
                                    });
          });

qPCR4vir:
This addition have no effect on any other feature or API

cnjinhao:
Thank you for your PR! Also thanks to beru's review.

PR #358, opened 2018-11-25
Add method to check option in group

JamesBremner:
The group widget has a method to get the checked status an option in the group.  However, there is no method to set the checked status.  This is needed when an application starts up so that it can set default values, or values from an pre-existing database that is to be edited.


JamesBremner:
Conflict is some extra whitespace in line 73 of group.hpp

cnjinhao:
Thank you a lot for your PR!

PR #359, opened 2018-11-28
Added tab_click event to tabbar control

besh81:
tab_click event is processed when a tab is clicked


besh81:
This PR add the tab_click event to tabbar control.

This event could be used to do different task based on the button clicked:
  - left mouse button: select tab
  - right mouse button: show contex menu
  - mid mouse button: close the tab

In addition the events of the tabbar (activated, removed, added and tab_click) now have the pos member that indicates the position of the clicked tab.

cnjinhao:
Thank you a lot for your PR!

PR #363, opened 2018-12-09
Prevent slider adorn moving when slider disabled

JamesBremner:
Fix suggested be Error Flynn in http://nanapro.org/en-us/forum/index.php?u=/topic/1059/ggslider-appearance-when-disabled#post-2498


cnjinhao:
Thank you a lot for this PR!

PR #369, opened 2018-12-14
fix disabled checkbox square

besh81:
disabled checkbox square now looks like disabled textbox
In addition found an unused instruction in text_editor.cpp: could be replaced or removed (or left as it is :) )


cnjinhao:
Thank you for your PR

Issue #362, opened 2018-12-09
static analysis fixes


Shamari91:
Hi,

I've been using Nana for some time for some personal projects and now i want to contribute to open source projects.

So was wondering If I could fix some static analysis issues ? The small coding issues found were from cppcheck.

Thanks,
Shamari

Shamari91:
Hi,

I have a small pull request for some cpp check issues.

It seems I don't have access to push any changes to my remote branch. (cloned from develop)

Would you be able to grant me access?

Thanks,

qPCR4vir:
Hi, @Shamari91,
just use GitHub to make a 'fork' of this repo. 
It will appear in your GitHub user profile. You will be able to push your commit there. And from there you will be able to make Pull Requests to this repo.
Read this please: [GitHub Standard Fork & Pull Request Workflow](https://gist.github.com/Chaser324/ce0505fbed06b947d962)

PR #370, opened 2018-12-16
Nana Issue #362

Shamari91:
Here are some small changes which I thought would help me get integrated with some of the code.

Below are some changes which i found after running CPPCheck over the solution also in some places I seen that some parameters could be declared const as they are not being changed.

https://github.com/cnjinhao/nana/issues/362


source/datetime.cpp
```diff
@@ -239,18 +240,20 @@ namespace nana
...
 
-		unsigned date::month_days(unsigned year, unsigned month)
+		unsigned date::month_days(const unsigned year, const unsigned month)
 		{
-			unsigned num[] = {31, 0, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
-			if(month != 2)
-				return num[month - 1];
+			if (month != 2)
+			{
+				const std::array<unsigned, 12> days_in_month = { 31, 0, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31 };
```
beru: How about using `constexpr` insted of `const` here?

Shamari91: Agreed, I have updated this patch with the change

cnjinhao:
Thank you for your PR

PR #387, opened 2019-02-06
 fixed listbox scheme.header_padding_top use 

besh81:


cnjinhao:
Thank you for fixing this bug!

PR #392, opened 2019-02-23
Some changes for Visual Studio project files

beru:
# .gitattributes
Added the file to specify line endings for Visual Studio solutions and projects files. Originally the line endings were `lf` but when Visual Studio updates the files, the application updates the line endings to `crlf`. Some git users use `core.autocrlf true` setting but actually that's not problem-free.

# Windows SDK Version
Not all Windows developers install specific version of Windows SDK (it was set to `10.0.14393.0`) so I specified `<inherit from parent or project defaults>` that resulted in `8.1` in my case.

image

# Multi-processor Compilation

Set project property Multi-processor Compilation to `Yes (/MP)` to shorten build time. I did the same thing in the past (#29) but somehow it's been nullified for some reason.

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+*.sln text eol=crlf
+*.vcxproj text eol=crlf
+*.vcxproj.filters text eol=crlf
+*.vcxproj.user text eol=crlf
+*.bat text eol=crlf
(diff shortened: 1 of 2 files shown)


cnjinhao:
Thanks for your PR!

Issue #408, opened 2019-03-18
linking problem related with std::(experimental::)filesystem with mingw-w64/gcc 7.3 on windows 10


crillion:
Hi,

   I am trying to compile a nana program with nana 1.7.0 and gcc 7.3.0 from mingw-w64 on windows 10. This is the program:

`
#include <nana/gui.hpp>

int main()
{
	nana::exec();
}
`

If I compile it with the command: 

`
g++ -pedantic -std=c++17 -Wall -Wextra -Werror=return-type -Wshadow=local -Wempty-body -fdiagnostics-color -s -Os program.cpp -o program_gpp.exe -LC:\programs\mingw-w64\x86_64-7.3.0-posix-seh-rt_v6-rev0\mingw64\lib -I../../../../external/nana/nana_1_7_0/nana/include -L../../../../external/nana/nana_1_7_0/nana -lnana -lgdi32 -lcomdlg32 -lstdc++fs -lKtmW32 -lws2_32
`

I get the error:

C:/programs/mingw-w64/x86_64-7.3.0-posix-seh-rt_v5-rev0/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/7.3.0/libstdc++fs.a(path.o):(.rdata$_ZNSt12experimental10filesystem2v17__cxx114path19preferred_separatorE+0x0): multiple definition of `std::experimental::filesystem::v1::__cxx11::path::preferred_separator'
../../../../external/nana/nana_1_7_0/nana/libnana.a(filebox.cpp.obj):filebox.cpp:(.rdata$_ZNSt12experimental10filesystem2v17__cxx114path19preferred_separatorE[_ZNSt12experimental10filesystem2v17__cxx114path19preferred_separatorE]+0x0): first defined here
collect2.exe: error: ld returned 1 exit status

If I remove from the compiling command the -lstdc++fs linking option, so that the compilation script becomes:

`
g++ -pedantic -std=c++17 -Wall -Wextra -Werror=return-type -Wshadow=local -Wempty-body -fdiagnostics-color -s -Os program.cpp -o program_gpp.exe -LC:\programs\mingw-w64\x86_64-7.3.0-posix-seh-rt_v6-rev0\mingw64\lib -I../../../../external/nana/nana_1_7_0/nana/include -L../../../../external/nana/nana_1_7_0/nana -lnana -lgdi32 -lcomdlg32 -lKtmW32 -lws2_32
`

 I get instead a long error thread:

[...]

PR #410, opened 2019-03-19
fix crash by using empty path to initialize folderbox with fs canonical() which don't exist

qPCR4vir:
After debugging, it resolved my problem in: https://github.com/cnjinhao/nana/issues/408

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,11 @@ cache:
 
 matrix:
   include:
-    - env: CXX=g++-5 CC=gcc-5
+    - env: CXX=g++-8
       addons:
         apt:
           packages:
-            - g++-5
+            - g++-8
             - libjpeg8-dev
             - libpng-dev
             - libasound2-dev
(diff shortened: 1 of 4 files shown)


qPCR4vir:
well, it seems that not many implementations of filesystem implement `weakly_canonical`. 
It may be better to catch the exception in `canonical` in `folderbox `?

cnjinhao:
Thanks for your PR!

qPCR4vir:
thank!
I updated my MS project to use the latest c++ standard and now I have `weakly_canonical`. 
Compiling OK with MinGW-gcc8.3 and MSVS2017

dudztroyer:
My VS2017 refuses to compile with ``weakly_canonical``
https://upload.wsbot.net/files/devenv_2019-04-08_01-44-32.png
``Not a member of std::filesystem`` and ``identifier not found`` are the translation for the errors.

qPCR4vir:
No problems, just read this:
https://stackoverflow.com/questions/50668814/vs2017-e0135-namespace-std-has-no-member-filesystem
Ask back if still have problems

PR #414, opened 2019-03-28
Added internationalization loading option with ifstream 

Wolowin:
Added internationalization loading option with ifstream instead of only string filename. Opens support for various usages - one of which being wstring paths on Windows.

### DIFF
--- a/include/nana/internationalization.hpp
+++ b/include/nana/internationalization.hpp
@@ -30,6 +30,9 @@ namespace nana
 		void load(const std::string& file);
 		void load_utf8(const std::string& file);
 
+		void load(std::ifstream& inputFileStream);
+		void load_utf8(std::ifstream& inputFileStream);
+
 		template<typename ...Args>
 		::std::string get(std::string msgid_utf8, Args&&... args) const
 		{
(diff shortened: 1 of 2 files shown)


cnjinhao:
Hi, Wolowin. Thank you for your PR!
IMO, it is not recommended to overload `internationalize::open()` for `std:;ifstream`. If a file is openned in text mode, it implies that all '\r\n" are translated to "\n", then it will create a result which is different from the original file.

Wolowin:
Hey,

I have to apologise for creating this merge request too fast - without making tripple sure that everything works. The fstream version that I suggested had some issue. I was in a rush at that day and I fell into "No need to check - it's a simple change, why wouldn't it work?" trap. Sorry for making you deal with this improfessionalmerge request. In the end I ended up using overload that accepts wstring instead of stream. I was low on time that day and wanted a solution as fast as possible. I will try to find a moment to make it work dependably for fstreams in a not-distant future.

Thank you for your time

cnjinhao:
Never mind, we can remove the std::string version and add a std::file system::path version that can open a utf-8/local encoding/wide string file name. Please feel free to contribute to this project. I’m looking forward to your next contribution.

Wolowin:
Hey! I have a question tho - does nana fully support utf-8 with both its internationalization class and dialogs?

PR #417, opened 2019-04-09
deprecated use of unsigned overloadings for functions in class timer

rbrugo:
Deprecated use of `unsigned` overloadings for `timer::timer` and `timer::interval`: as mentioned in #416 `std::chrono::duration`s should be preferred over `unsigned int`s to represent time intervals.
I preferred to deprecate the features instead of removing them (as proposed in #416) as this would have led to breaking changes.

(I didn't have deprecated `unsigned timer::interval() const`; I can do it if needed)

### DIFF
--- a/include/nana/gui/timer.hpp
+++ b/include/nana/gui/timer.hpp
@@ -38,7 +38,9 @@ namespace nana
 		timer& operator=(timer&&) = delete;
 	public:
 		timer();
-		timer(unsigned int ms) : timer{} { interval(ms); } /// Accepts an initial interval in ms
+		[[deprecated("prefer a std::chrono::duration (like std::chrono::milliseconds) over "
+			     "raw integers for durations")]]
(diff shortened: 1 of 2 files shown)


cnjinhao:
Hi, thank you for your PR! Now I am merging this PR, but before I commit it. I think we should change the template function
```
//from
template <typename Rep, typename Period>
inline void interval(std::chrono::duration<Rep, Period> const & time_interval);

//to
void interval(std::chrono::milliseconds ms);
```

Because the smallest uint of timer interval is millisecond. If a programmer wants to set 100-millisecond for interval, but he mistakely passes a nanoseconds, e.g.`timer.interval(std::chrono::nanoseconds{100});`, the template function allows nanoseconds and it results a 0 millisecond.

The second version of interval, it will cause a compiling error to tell the programmer that a wrong parameter is passed.

PR #429, opened 2019-05-05
two listbox bug fixes: category_icon() parameter and an unsort issue

ErrorFlynn:


cnjinhao:
Thank you for fixing this bug!

PR #430, opened 2019-05-08
toolbar : added toggle button

besh81:
I've worked a bit on the toolbar widget.
The main addition are toggle button (with group management).
A toggle button can assume 2 different state: not pushed (default) or pushed. Clicking on a toggle button it change from not pushed to pushed and viceversa.
A toggle button can be added to a group. These are the group rules:
   1. inside a group only one item at the time is pushed
   2. inside a group one item must always be pushed
   3. a group with only one item IS NOT a group

Following functions/methods have been add both to toolbar and item_proxy:
bool istoggle() const; ///< Returns true if the tool style is toggle.
				bool toggle() const; ///< Gets the tool toggle state (only if tool style is toggle).
				item_proxy& toggle(bool toggle_state); ///< Sets the tool toggle state (only if tool style is toggle).
				std::string toggle_group() const;	///< Returns the toggle group associated with the tool (only if tool style is toggle).
				item_proxy& toggle_group(const ::std::string& group);	///< Adds the tool to a toggle group (only if tool style is toggle).

Ciao

### DIFF
--- a/include/nana/gui/widgets/toolbar.hpp
+++ b/include/nana/gui/widgets/toolbar.hpp
@@ -55,6 +55,8 @@ namespace nana
 				std::string toggle_group() const;	///< Returns the toggle group associated with the tool (only if tool style is toggle).
 				item_proxy& toggle_group(const ::std::string& group);	///< Adds the tool to a toggle group (only if tool style is toggle).
 
+				item_proxy& textout(bool show); ///< Show/Hide the text inside the button
+
 			private:
 				nana::toolbar* const tb_;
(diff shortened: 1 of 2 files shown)


besh81:
I've also added function/method to show text inside the button and changed the display behavior of buttons without images.

Added function/method both to toolbar and item_proxy:
void textout(size_type index, bool show); ///< Show/Hide the text inside the button

Before: button without image weren't display
Now: button without image is display as empty square. In case the textout is on the text is show

cnjinhao:
Thank you for your PR!

PR #437, opened 2019-05-15
c++ runtime static linking

qPCR4vir:
This prevent the need to copy some dll with some MinGW distribution.
Tested with gcc 9.1.1, the lastest optionfrom https://gcc-mcf.lhmouse.com/
 
It set: CMAKE_CXX_COMPILER_VERSION = 9.1.1 !!!

I changed the build/cmake/compilers.cmake to add static c++ runtime linking but still need to copy mcfgthread-12.dll with the .exe files.
How to fix that? I want all static linked, and move around just my .exe file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,27 +54,25 @@ target_compile_features(nana
 # in your own CMakeLists.txt, and them :
 # target_link_libraries(yourApp PRIVATE nana )
 
-set(NANA_SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/source)
-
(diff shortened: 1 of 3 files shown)


qPCR4vir:
Now travis errors: see https://cmake.org/cmake/help/latest/command/target_link_libraries.html?highlight=target_link_libraries
How correctly add the links option?

Xeverous:
It's currently impossible to statically link MCF thread implementation in the compiler built by LH_Mouse. See https://github.com/lhmouse/mcfgthread/wiki#distribution-notes for explanation

cnjinhao:
Thank you for the PR!

Issue #440, opened 2019-05-22
unicode_bidi aletter always true


glavangeorge:
```
		if (('A' <= ch && ch <= 'Z') || ('a' <= ch && ch <= 'z') || (0x00AA == ch || 0x00B5 == ch || 0x00BA == ch) || (0x00C0 <= ch && ch <= 0x00D6) ||
			(0x00D8 <= ch && ch <= 0x00F6) || (0x00F8 <= ch && ch <= 0x0236) || (0x0250 <= ch || ch <= 0x02C1))
return unicode_character_type::aletter;
```

from here [https://github.com/cnjinhao/nana/blob/master/source/unicode_bidi.cpp#L979](url). 

shouldn't it be   (0x0250 <= ch && ch <= 0x02C1) instead of  (0x0250 <= ch || ch <= 0x02C1)?

cnjinhao:
Thank you for pointing out this bug!

PR #441, opened 2019-05-23
fix for #440 (unicode_bidi aletter always true)

glavangeorge:


cnjinhao:
Thanks a lot for fixing this bug!

Issue #448, opened 2019-06-03
Listbox selection bug


ErrorFlynn:
This bug is a little hard to describe, so I'll just describe some steps that can be used to reproduce it with the test code below:
1. click `item abs3`
2. shift-click `item abs7`
3. click `item abs5`
4. shift-click `item abs6`

The last step should select items abs5 and abs6, but instead it selects the range from 0 to 6.

Test code:
```
#include <nana/gui.hpp>
#include <nana/gui/widgets/listbox.hpp>

int main()
{
	using namespace nana;

	form fm{API::make_center(345, 678)};
	fm.div("vert margin=15 <listbox>");

	listbox lb{fm};
	fm["listbox"] << lb;
	
	lb.append_header("asdads");
	for(int n(0); n<16; n++)
		lb.at(0).append("item abs" + std::to_string(n));

	fm.collocate();
	lb.column_at(0).width(lb.size().width-16);
	fm.show();
	nana::exec();
}
```

I will submit a PR with the fix.

PR #449, opened 2019-06-03
fix for listbox selection bug described in issue #448

ErrorFlynn:


cnjinhao:
Thank you for your PR!

Issue #423, opened 2019-04-22
Linking to stdc++fs.lib fails


khrynczenko:
I tried building my project with nana with latest commit (commit ffa170b5f560dbf899030bdeb0e539aa95b0364) and was unable to do so beacuse of linkage error:
`LNK1104	cannot open file 'stdc++fs.lib'`
I am using Visual Studio 15 2017 version 15.9.11.

cnjinhao:
There is not a stdc++fs file in Visual Studio. Please remove the linker option.

khrynczenko:
Yes, I did that already. My point is why CMake generates project with such linker option in the first place.

khrynczenko:
Okay so my question is do we really need to link `stdc++fs` explicitly in file `select_filesystem.cmake`? We already define `target_compile_features(nana PUBLIC cxx_std_17)` in `CMakeLists.txt` so is it really necessary? This linking part breaks my builds on Windows.  
https://github.com/cnjinhao/nana/blob/39c3f85627571e9d3420781787a576ce8e300ac1/build/cmake/select_filesystem.cmake#L46  
https://github.com/cnjinhao/nana/blob/39c3f85627571e9d3420781787a576ce8e300ac1/CMakeLists.txt#L33

5cript:
Yes since you can have C++17 enabled without using std::filesystem on g++.
And I think clang doesn't need the linker line too.

This is one rare case where checking against the compiler is a viable method:
https://stackoverflow.com/a/50173063

khrynczenko:
@5cript Thank you for answering. Indeed in order to use std::filesystem with GCC stdc++fs needs to be linked explicitly. I found out that in newest GCC release (9.1) it is not longer required to do that but yeah I suppose we still want to be able to build *nana* on older versions. Anyway I am going to add PR with changes so that stdc++fs is only linked on GNU using systems.
![obraz](https://user-images.githubusercontent.com/15171153/59976232-d073fd80-95c1-11e9-9023-930554c5f9c7.png)

qPCR4vir:
Solved by https://github.com/cnjinhao/nana/pull/494

PR #453, opened 2019-06-23
Stopped linking stdc++fs on windows

khrynczenko:
Please refer to #423 for more details.  
Added conditions in `select_filesystem.cmake` code so that when building on Windows `stdc++fs` is not being passed to linker.

### DIFF
--- a/build/cmake/select_filesystem.cmake
+++ b/build/cmake/select_filesystem.cmake
@@ -21,7 +21,9 @@ if(NANA_CMAKE_NANA_FILESYSTEM_FORCE)
 
 elseif(NANA_CMAKE_STD_FILESYSTEM_FORCE)
     target_compile_definitions(nana PUBLIC STD_FILESYSTEM_FORCE)
-    target_link_libraries     (nana PUBLIC stdc++fs)
+	if (CMAKE_CXX_COMPILER_ID STREQUAL GNU)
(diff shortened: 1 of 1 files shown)


khrynczenko:
@qPCR4vir what about this approach now?  
And as to your question. For my project I use highly automated CMake script where nana is one of many dependencies. I want to be cross-platform as possible so I don't want to use Visual Studio project only. Anyway basically I have something like this.
```
FetchContent_Declare(
        nana
        GIT_REPOSITORY "https://github.com/cnjinhao/nana.git"
        GIT_TAG "v1.7.1"
        GIT_SHALLOW TRUE
        GIT_PROGRESS TRUE
        UPDATE_COMMAND ""
        UPDATE_DISCONNECTED TRUE
        INSTALL_COMMAND ""
        TEST_COMMAND ""
)
# many more such declarations ...

add_subdirectory(${nana_SOURCE_DIR} EXCLUDE_FROM_ALL)
# many more ...
```

khrynczenko:
What's wrong with this pull request? With some help and feedback I could maybe improve it.

cnjinhao:
Thank you for your effort. As @qPCR4vir mentioned, clang should be considered(and libc++fs). Even though all compilers had been supported by the cmake script, we still need take care about the versions of compilers. A newer version clang doesn’t need the dependency of libc++fs

PR #461, opened 2019-07-27
 missing target_include_directories in CMakeLists.txt

Xeverous:
Nana lacked exposition of public headers. Any target which links to nana (by target_link_libraries(project_name PRIVATE nana)) was failing to compile because nana did not expose public headers (#include <nana/...> were not found). 

This change adds a missing target_include_directories() command which will cause targets that link to nana to get nana's include directory.


qPCR4vir:
No, this will be an error if you want to install nana. 
Please see: https://github.com/cnjinhao/nana/blob/master/build/cmake/install_nana.cmake#L21
Try to see why you have this problem.

qPCR4vir:
please test:
https://github.com/cnjinhao/nana/pull/463

Xeverous:
Ok, I understand. #463 is the correct way to do this. Will try and report.

qPCR4vir:
hi, thank! This can be closed now

Xeverous:
Tested. The build works.

PR #463, opened 2019-07-28
FIX 4. inputbox buttons overlaid, missed included in installed nana and both travis and appveyor IC

qPCR4vir:
from https://cmake.org/cmake/help/v3.0/command/target_include_directories.html

> Include directories usage requirements commonly differ between the build-tree and the install-tree. The BUILD_INTERFACE and INSTALL_INTERFACE generator expressions can be used to describe separate usage requirements based on the usage location. Relative paths are allowed within the INSTALL_INTERFACE expression and are interpreted relative to the installation prefix. For example:

```
target_include_directories(mylib PUBLIC
  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/mylib>
  $<INSTALL_INTERFACE:include/mylib>  # <prefix>/include/mylib
)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,14 +53,14 @@ install:
   - /tmp/tools/cmake --prefix="$HOME" --exclude-subdir
 
 before_script :
-  # travis don't have a physical monitor. We need to install an emulator: https://docs.travis-ci.com/user/gui-and-headless-browsers/
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-  - sleep 3 # give xvfb some time to start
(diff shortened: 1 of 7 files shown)


Xeverous:
This works, no more include errors.

You should not commit commented code though. There is already a lot of commented code in cmake files.

qPCR4vir:
[Travis CI XVFB Breaking Change](https://benlimmer.com/2019/01/14/travis-ci-xvfb/)

Xeverous:
Use `#define NOMINMAX` just before including `windows.h`, you may also need to add `#define WIN32_LEAN_AND_MEAN`.

cnjinhao:
Thanks for your PR!

Issue #445, opened 2019-05-29
crash on Windows upon startup


Xeverous:
Hello, I'm trying to run any of the examples but all crash on startup.

Minimal code:

```c++
#include <nana/gui/wvl.hpp>
#include <nana/gui/widgets/label.hpp>
int main()
{
    using namespace nana;
    form    fm;
    label   lb(fm, rectangle(fm.size()));
    lb.caption("Hello, World");
    fm.show();
    exec();
}
```

GDB backtrace:

```
#0  0x00007ffd430fe912 in ntdll!RtlCallEnclaveReturn ()
   from C:\Windows\SYSTEM32\ntdll.dll
#1  0x00007ffd4306bbce in ntdll!RtlRaiseException ()
   from C:\Windows\SYSTEM32\ntdll.dll
#2  0x00007ffd3f9aa388 in RaiseException ()
   from C:\Windows\System32\KernelBase.dll
#3  0x000000006144d711 in ?? ()
   from [...] libgcc_s_seh-1.dll
#4  0x000000006fd070c5 in ?? ()
   from [...] libstdc++-6.dll
#5  0x000000006fcfb1d7 in ?? ()
   from [...] libstdc++-6.dll
#6  0x0000000000405694 in void std::__cxx11::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> >::_M_construct<wchar_t const*>(wchar_t const*, wchar_t const*, std::forward_iterator_tag) [clone .constprop.292] ()
#7  0x000000000040a512 in nana::to_utf8[abi:cxx11](std::basic_string_view<wchar_t, std::char_traits<wchar_t> >) ()
#8  0x000000000040a981 in nana::font_factory(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, double, nana::detail::font_style const&, std::filesystem::__cxx11::path) ()
#9  0x0000000000110bfa in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```

qPCR4vir:
exactly what nana version or commit are you using? How are you compiling?

Xeverous:
Just checked and pulled newest code, it works on `742cde9779d74e4bdd9bff293b0004129d50f700` but was crashing on `ffa170b5f560dbf899030bdeb0e539aa95b03649`. Now, after full rebuild of nana I can't reproduce it.

Xeverous:
[...]

PR #467, opened 2019-07-31
platform_runtime - fixed static initialization bug

Xeverous:
platform_runtime was not guuaranteed to be initialized. Changed
implementation to Meyer's singleton. Any other file which needs
platform_runtime must include platform_abstraction.hpp and use
platform_runtime_instance(). Calling the function will prevent
the linker from delaying or removing initialization.

### DIFF
--- a/source/detail/platform_abstraction.cpp
+++ b/source/detail/platform_abstraction.cpp
@@ -511,6 +511,7 @@ namespace nana
 
 	struct platform_runtime
 	{
+
 		std::shared_ptr<font_interface> font;
 
 #ifdef NANA_X11
@@ -521,22 +522,11 @@ namespace nana
 #endif
 	};
 
-	namespace
+	platform_runtime& platform_runtime_instance()
 	{
-		namespace data
-		{
-			static platform_runtime* storage;
-		}
+		static platform_runtime runtime;
+		return runtime;
 	}
-
-	static platform_runtime& platform_storage()
-	{
-		if (nullptr == data::storage)
-			throw std::runtime_error("platform_abstraction is empty");
-
-		return *data::storage;
-	}
-
 
 	class internal_font
 		: public font_interface
@@ -570,7 +560,7 @@ namespace nana
 #elif defined(NANA_X11)
 			auto disp = ::nana::detail::platform_spec::instance().open_display();
 #	ifdef NANA_USE_XFT
-			platform_storage().fb_manager.release_fallback(fallback_);
+			platform_runtime_instance().fb_manager.release_fallback(fallback_);
 			::XftFontClose(disp, reinterpret_cast<XftFont*>(native_handle_));
 #	else
 			::XFreeFontSet(disp, reinterpret_cast<XFontSet>(native_handle_));
@@ -650,26 +640,6 @@ namespace nana
 	}
 #endif
 
-
-	void platform_abstraction::initialize()
-	{
-		if (nullptr == data::storage)
-			data::storage = new platform_runtime;
-	}
-
-	void platform_abstraction::shutdown()
-	{
-		auto & r = platform_storage();
-
-		if (r.font.use_count() > 1)
-			throw std::runtime_error("platform_abstraction is disallowed to shutdown");
-
-		r.font.reset();
-
-		delete data::storage;
-		data::storage = nullptr;
-	}
-
 	double platform_abstraction::font_default_pt()
 	{
 #ifdef NANA_WINDOWS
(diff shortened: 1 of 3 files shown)


Xeverous:
This fixes one of the crash problems of #445.

Xeverous:
Something is wrong with CI.

qPCR4vir:
> Something is wrong with CI.

Yes, Already fixed in #463. It will be solved after merging #463. You may cherry pick  commit [not wanting to start xfvb](https://github.com/cnjinhao/nana/pull/463/commits/bcf604a9b608ee9cbd2339772eda82216791d072)

Xeverous:
Can not cherry-pick commits from different repository. Will wait for merge of #463 and then rebase this PR.

source/detail/platform_spec_windows.cpp
```diff
@@ -114,12 +114,11 @@ namespace detail
 	platform_spec::platform_spec()
 		: impl_{ new implementation}
 	{
-		platform_abstraction::initialize();
+		platform_runtime_instance(); // force initialization
 	}
 
 	platform_spec::~platform_spec()
 	{
-		platform_abstraction::shutdown();
```
cnjinhao: This line is removed, because platform_runtime runtime now is a static object. The problem is that we can control the life time of the object.

The code below is the implementation of `platform_spec` in Linux
```
	platform_spec::~platform_spec()
	{
		delete msg_dispatcher_;

		//The font should be destroyed before closing display,
		//otherwise it crashs
		platform_abstraction::shutdown();

		close_display();
	}
```
We must destroy the `platform_runtime` before closing X display.

source/detail/platform_abstraction.cpp
```diff
@@ -650,26 +640,6 @@ namespace nana
...
-
-	void platform_abstraction::initialize()
-	{
-		if (nullptr == data::storage)
-			data::storage = new platform_runtime;
-	}
-
-	void platform_abstraction::shutdown()
-	{
-		auto & r = platform_storage();
```
cnjinhao: A check is needed before destroying the `platform_runtime`. Otherwise, when the program exists, it generates a incomprehensible crash if there is still a font object not releasing

cnjinhao:
Hi, @Xeverous ,
Thanks for your PR, I reviewed the changes and noticed some issues.

source/detail/platform_spec_windows.cpp
Xeverous: 1. I actually forgot to edit `platform_spec_posix.cpp`, right now the code will not compile.

2. In both cases (Windows, POSIX) we want `platform_spec` object to outlive `platform_runtime` object. We can easily guuarantee this by making `platform_runtime` a member of `platform_spec`.

We should not write things like `foobar::initialize()`/`foobar::shutdown()` because that's not really modern C++. It's like `new`/`delete`. What we should do instead, is to put `platform_runtime` inside `platform_spec` and the correct order of construction/destruction is guuaranteed.

Will edit appropriately.

source/detail/platform_abstraction.cpp
Xeverous: What we should do if the check fails? Previously an exception was thrown ... but throwing in `platform_abstraction::shutdown()` which is called in `platform_spec` destructor is also a crash.

It would be better to guuarantee that all objects which used `font` are destroyed before `platform_runtime` is destroyed.

Xeverous:
Hey, you still interested in this PR?

Few things left unanswered:

- Do you agree to make `platform_runtime` a member of `platform_spec`? This trivializes construction and destruction and guuarantees correct order of operations.
- What to do with font usage?
  - Right now: you check for font usage and if it is used you throw in a destructor => crash
  - My proposition: no check, just let destructor release the font. We can guuarantee no font usage through  correct order of destruction of everything that uses fonts. If the guuarantee fails, it will still crash but the code is simpler and easier to fix remaining crashes.

cnjinhao:
>Do you agree to make platform_runtime a member of platform_spec? This trivializes construction and destruction and guuarantees correct order of operations.

Agree. BTW. platform_runtime is the "data member" of `platform_abstraction`.  `platform_abstraction` provides os-independent APIs, `platform_spec` provide os-specific APIs.

>What to do with font usage? 

```
	void platform_abstraction::shutdown()
	{
		auto & r = platform_storage();

		if (r.font.use_count() > 1)
			throw std::runtime_error("platform_abstraction is disallowed to shutdown");

		r.font.reset(); //explicitly release the font object.

		//...
	}
```
All the font objects must be destructed before `platform_spec` is destructed, the reason that in X(Linux), the font objects must be destructed before closing X display, otherwise, when closing X display, a crash occurs. But the crash is a general X error, it is hard to understand why the crash occured. So I added the check to tell programmers if a font object is leaked. 

The problem is that the check not always works, because it only checks the use count of the default font.

Xeverous:
So what do we do with the font? Leak and log it?

> font objects must be destructed before closing X display, otherwise, when closing X display, a crash occurs

I understand this, which is a bit sad because it creates some hidden invariants: that is, you can not independently create/destry font and display objects. How does X connect them? I guess nana is calling some X functions when creating fonts and displays?

This problem appears to be larger than I thought so I will likely need to read the implementation of fonts and how they interact with displays. I would prefer to find a way which allows independent creation/destruction of font and display objects.

PR #468, opened 2019-08-01
fix the strange behavior of Korean ime composition window.

JannLee:
Because Hangul is a combination character, WM_IME_STARTCOMPOSITION is only called once when the IME input is started, so the position of the composition window is strange.

Therefore, I solved the problem by controlling the state of characters combined using WM_IME_COMPOSITION and WM_IME_CHAR.

We have also improved to support other IME languages such as Chinese and Japanese.


cnjinhao:
Thank you a lot for the PR!

PR #469, opened 2019-08-01
added option to link C++ standard library statically

Xeverous:
Note: I'm not sure whether this default is the right choice (most projects I have seen used dynamic by default) but at least this commit gives a choice to do otherwise.

I'm not even sure if such thing should be in the CMake scripts at all - perhaps we should do nothing in regards to standard libraries and let everyone use their compiler's default or global CMake settings.


Xeverous:
This is another fix for #445 problems.

cnjinhao:
Thanks for your PR!

PR #477, opened 2019-08-11
On macOS and FreeBSD libc++ is the default standard library and the -…

loaden:
…stdlib=libc++ is not required.

https://libcxx.llvm.org/docs/UsingLibcxx.html


cnjinhao:
Thank you for your PR!

PR #482, opened 2019-09-01
Fix inconsistency regarding filebox multi_select

mihaizh:
When only one file is selected, the `lpstrFile` variable contains only the path to that file, without the parent path before that, thus, the `targets` vector is empty and the `path` variable contains the actual path to the file. This is inconsistent with the scenario where multiple files are selected, in which `targets` contains the file paths and `path` contains the parent directory path.


cnjinhao:
Thank you for fixing this issue! This PR has been merged into `hotfix-1.7.2` branch.

PR #483, opened 2019-09-05
dock_pane and toolbar bug fix

besh81:
fixed bug in dock_pane caption drawing function (used a workaround that hide the caption when the dock_pane width become too small)
fixed small bug in toolbar


cnjinhao:
Thanks for your PR!

PR #487, opened 2019-09-14
Bug fixes for listbox, treebox, and nana::any

ErrorFlynn:
- `treebox.cpp`:  fixed a bug in `trigger::dbl_click()`
`impl_->set_expanded()` was called unconditionally, even if the node had no children. This caused the node icon to change to the "expanded" icon if the node had an icon scheme, even when the node didn't have children.
- `listbox.cpp`:  `listbox::sort_col` doesn't update view
Calling `listbox::sort_col` to change the sort column doesn't update the viewport to reflect the change, forcing the user to call `API::refresh_window` or otherwise perform stupid tricks to force the listbox to refresh.
- `any.hpp`: argument not forwarded
A constructor and an overload of the assignment operator each have a forwarding reference as a parameter, but they don't actually forward the argument.
- `treebox.cpp`:  fixed buggy `item_proxy` postfix increment operator
The behavior of the postfix increment operator is not consistent with the increment operator concept (currently just returns the next sibling node).
- `listbox.cpp`: `nana::drawerbase::listbox::essence::where` incorrectly calculates the position of checkboxes in the listbox content area (see the extensive description in the commit)

### DIFF
--- a/include/nana/any.hpp
+++ b/include/nana/any.hpp
@@ -74,7 +74,7 @@ namespace nana
 		any(Value && value,
 				typename std::enable_if<!std::is_same<any&, Value>::value>::type * = nullptr,
 				typename std::enable_if<!std::is_const<Value>::value>::type* = nullptr)
-			: content_(new holder<typename std::decay<Value>::type>(static_cast<Value&&>(value)))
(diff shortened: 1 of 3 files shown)


cnjinhao:
Thank you for fixing this issue.

PR #488, opened 2019-09-15
Fix tabbar drawing not updated when a tab is erased

mihaizh:
When a tab is deleted using `tabbar::erase`, `trigger::erase` is called. This function does not call `API::refresh_window` if the tab was erased successfully (unlike `trigger::tab_color`, `trigger::tab_image`, etc.), thus the tab is still visible until an interaction is made (like mouse click on `tabbar`).


cnjinhao:
Thank you for your PR!

PR #492, opened 2019-10-04
fix static stdlib generator expression

ppetraki:
The OR syntax doesn't work.

```
[100%] Linking CXX shared library libnana.so
c++: error: $<$<OR:$: No such file or directory
c++: error: unrecognized command line option ‘-static-libstdc++>’; did you mean ‘-static-libstdc++’?
CMakeFiles/nana.dir/build.make:1260: recipe for target 'libnana.so' failed
make[2]: *** [libnana.so] Error 1
CMakeFiles/Makefile2:75: recipe for target 'CMakeFiles/nana.dir/all' failed
make[1]: *** [CMakeFiles/nana.dir/all] Error 2
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
```

Only one COMPILER_ID can be true at any one time. The rest will evaluate to nothing.

### DIFF
--- a/build/cmake/shared_libs.cmake
+++ b/build/cmake/shared_libs.cmake
@@ -38,6 +38,9 @@ endif()
 
 if(NANA_STATIC_STDLIB)
     target_link_libraries(nana
-        PUBLIC $<$<OR:$<CXX_COMPILER_ID:GNU>, $<CXX_COMPILER_ID:Clang>>: -static-libgcc -static-libstdc++>)
+        PUBLIC
+            $<$<CXX_COMPILER_ID:GNU>:-static-libgcc>
+            $<$<CXX_COMPILER_ID:Clang>:-static-libstdc++>
+            )
 endif()
 


cnjinhao:
Thank you for your PR!

qPCR4vir:
It was very unfortunate that we accepted #469 from @Xeverous : added option to link C++ standard library statically
 
See: #474  FIX reversion in static linking introduced in #469

Now we have another regression.

ppetraki:
Perhaps the entire customization point is over engineered? Had the original feature been introduced with CI support that stressed that configuration path this simply wouldn't have happened to begin with.

And uh... libgcc is not a C++ library, I missed that, and while I fixed the syntax so the project could be configured, now we have something that doesn't link. So my bugfix just exaggerated a bug that already existed.

Not only is the cmake generator logic obtuse, even if the syntax was correct to begin with, it would introduce problems on [Mac platforms](https://stackoverflow.com/questions/19774778/when-is-it-necessary-to-use-use-the-flag-stdlib-libstdc). The use of g++ doesn't mean it's using g++ or using the "default" c++ library associated with said compiler.

Also "static" doesn't really mean pure static when you deal with [libc](https://stackoverflow.com/questions/26304531/compiling-with-static-libgcc-static-libstdc-still-results-in-dynamic-depende).

> GNU libc is not designed to be statically linked. Important functions, e.g. gethostbyname and iconv, will malfunction or not work at all in a static binary. Arguably even worse, under some conditions a static binary will attempt to dynamically open and use libc.so.6, even though the whole point of static linkage is to avoid such dependencies.
> 
> You should compile your program against uClibc or musl libc instead.
> 
> (This has been true for at least 15 years.)

Perhaps removing the static linking option is the best option until an end to end PR can address and test (travis CI support) the static linking process (and the definition of "static" e.g. dlopen libc is OK or another portable static libc must be used ) on a per platform and compiler option basis?

This issue was originally discovered in the pursuit of packaging Nana for Conan. Once you're in this pipeline you must be able to support static and shared configuration options and different compilers. The nice thing about targeting Conan is they have already developed a nice CI ecosystem around this to catch this class of bugs. Without that or similar infrastructure this is going to keep happening.

Xeverous:
I'm actually surprised of the `-static-libstdc++>`. How it did not make a CMake syntax error?

Also, the `-static-libgcc -static-libstdc++` flags were originally meant to be for both GCC and Clang compilers. That is, add both `-static-libgcc -static-libstdc++` if the compiler is GCC or Clang.

ppetraki:
@Xeverous  Who knows honestly. I actually tried to reproduce the error that prompted my original patch and was unsuccessful (newer cmake?). Worst, how do you actually verify it's doing the right thing? If you try to inspect the [target's attributes ](https://stackoverflow.com/questions/32183975/how-to-print-all-the-properties-of-a-target-in-cmake) all you get to see is the generator expression is inlined into the LINK_ variables. But if you actually try to find the flags in any of the support scripts cmake generates it is nowhere to be found

<PRE>
nana INTERFACE_LINK_LIBRARIES = /usr/lib/x86_64-linux-gnu/libSM.so;/usr/lib/x86_64-linux-gnu/libICE.so;/usr/lib/x86_64-linux-gnu/libX11.so;/usr/lib/x86_64-linux-gnu/libXext.so;/usr/lib/x86_64-linux-gnu/libXft.so;/usr/lib/x86_64-linux-gnu/libfreetype.so;/usr/lib/x86_64-linux-gnu/libfontconfig.so;$<$<CXX_COMPILER_ID:GNU,Clang>:-static-libgcc;-static-libstdc++>;$<LINK_ONLY:Threads::Threads>

nana LINK_LIBRARIES = /usr/lib/x86_64-linux-gnu/libSM.so;/usr/lib/x86_64-linux-gnu/libICE.so;/usr/lib/x86_64-linux-gnu/libX11.so;/usr/lib/x86_64-linux-gnu/libXext.so;/usr/lib/x86_64-linux-gnu/libXft.so;/usr/lib/x86_64-linux-gnu/libfreetype.so;/usr/lib/x86_64-linux-gnu/libfontconfig.so;$<$<CXX_COMPILER_ID:GNU,Clang>:-static-libgcc;-static-libstdc++>;Threads::Threads
</PRE>

Call me paranoid, but I'd like to see the flag somewhere in the build scripting. This isn't encouraging.
NOTE: The formatter is devouring the inner generator expression which is "$<$<CXX_COMPILER_ID:GNU,Clang>:-static-libgcc -static-libstdc++>"

<PRE>
ppetraki@vanguard:~/Sandbox/hacking/nana-develop$ git diff
...
 
-if(NANA_STATIC_STDLIB)
-    target_link_libraries(nana
-        PUBLIC
-          $<$<CXX_COMPILER_ID:GNU,Clang>:-static-libgcc -static-libstdc++>
-          )
-endif()
+#if(NANA_STATIC_STDLIB)
+#    target_link_libraries(nana
+#        PUBLIC
+#          $<$<CXX_COMPILER_ID:GNU,Clang>:-static-libgcc -static-libstdc++>
+#          )
+#endif()
</PRE>

<PRE>
ppetraki@vanguard:~/Sandbox/hacking/nana-develop$ ls -l Release/libnana.a
-rw-r--r-- 1 ppetraki ppetraki 9066382 Dec  5 16:17 Release/libnana.a
ppetraki@vanguard:~/Sandbox/hacking/nana-develop$ ls -l Release2/libnana.a
-rw-r--r-- 1 ppetraki ppetraki 9066382 Dec  5 16:19 Release2/libnana.a
</PRE>

It doesn't look like those flags are having any impact at all.

ppetraki:
Ah, You'll never see it because it's part of the contract passed to users of the nana library when it's configured in this way. Carry on. It could probably use a regression test though to ensure libstdc++ is not being dynamically linked.

PR #496, opened 2019-12-05
 FIX travis 

qPCR4vir:
This will FIX travis broken by #469, #492
Run-test one more demo: helloworld_demo
Improve performance of listbox
Run all test

### DIFF
--- a/build/cmake/shared_libs.cmake
+++ b/build/cmake/shared_libs.cmake
@@ -39,8 +39,8 @@ endif()
 if(NANA_STATIC_STDLIB)
     target_link_libraries(nana
         PUBLIC
-            $<$<CXX_COMPILER_ID:GNU>:-static-libgcc>
-            $<$<CXX_COMPILER_ID:Clang>:-static-libstdc++>
+            $<$<CXX_COMPILER_ID:GNU>:-static-libgcc -static-libstdc++>
+            $<$<CXX_COMPILER_ID:Clang>:-static-libgcc -static-libstdc++>
             )
 endif()
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,8 +43,12 @@ matrix:
             - llvm-toolchain-precise 
 
 before_install:
+  # donwload nana-demo first
+  # we are in:  'user'/nana/
   - cd ..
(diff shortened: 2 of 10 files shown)


build/cmake/shared_libs.cmake
```diff
@@ -39,8 +39,8 @@ endif()
 if(NANA_STATIC_STDLIB)
     target_link_libraries(nana
         PUBLIC
-            $<$<CXX_COMPILER_ID:GNU>:-static-libgcc>
-            $<$<CXX_COMPILER_ID:Clang>:-static-libstdc++>
+            $<$<CXX_COMPILER_ID:GNU>:-static-libgcc -static-libstdc++>
+            $<$<CXX_COMPILER_ID:Clang>:-static-libgcc -static-libstdc++>
```
ppetraki: Recently I learned the following syntax accomplishes the same thing:
`$<$<CXX_COMPILER_ID:GNU,Clang>:-static-libgcc -static-libstdc++> `

See, https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

> $<CXX_COMPILER_ID:compiler_ids>
> where compiler_ids is a comma-separated list. 1 if the CMake’s compiler id of the CXX compiler matches any one of the entries in compiler_ids, otherwise 0. See also the CMAKE_<LANG>_COMPILER_ID variable.

qPCR4vir: Nice to see. Our CI travis tested on Ubuntu 16 - xenial with gcc 8. 
Do you have experience running travis with macos or at least with clang?

ppetraki: Some. Mostly by reusing the docker containers Conan has already vetted. See, https://github.com/ppetraki/conan-sdl2pp/blob/master/.travis.yml

Those osx containers work. I just haven't gotten to supporting that platform yet. The rest is just boilerplate lifted from other conan example packages e.g. .ci/ and build.py .

ppetraki: You might find this useful.
https://github.com/conan-io/conan-docker-tools

cnjinhao:
Thank you for the PR!

PR #500, opened 2020-01-22
treebox - allow treebox node to be hidden

call-me-void:
Treebox items can be hidden and shown. It is useful feature for filtering or searching on large treeboxes, as it's now possible to work with the part of the treebox without actually removing any items.

Implementation mostly inspired by the implementation of the "expanded" property.


cnjinhao:
Thank you a lot for this PR!

PR #502, opened 2020-02-03
fix: submenu option items unresponsive to clicks

ErrorFlynn:
When the user opens a submenu containing items with the style `menu::checks::option` and clicks an option, the radio button logic doesn't work. This bug was discovered and fixed by nanapro.org forum user "huycan" (I'm just submitting the PR). See this forum thread: http://nanapro.org/en-us/forum/index.php?u=/topic/1256/ggleave-a-menu-open

As mentioned in more detail in the forum thread linked above, my understanding of why this happens is that `menu_builder::checked` fails because the `menu_window` of the submenu is created with the `menu_builder` object used by the parent menu.

### DIFF
--- a/source/gui/widgets/menu.cpp
+++ b/source/gui/widgets/menu.cpp
@@ -921,7 +921,25 @@ namespace nana
 					}
 					else if (checks::option == item.style)
 					{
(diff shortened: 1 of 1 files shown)


cnjinhao:
Thank you a lot for this fix.

PR #504, opened 2020-02-07
added treebox use_entire_line mode

besh81:
Added new interaction/display mode to treebox control.
Enabling use_entire_line the entire line of the treebox/child is used for mouse interactions and display purposes.


cnjinhao:
Thank you a lot for this new feature!

PR #509, opened 2020-02-17
Include<stdexcept> in datetime.cpp

LilyWangL:
In a future version of MSVC, <string> doesn't transitively include<ostream>.
This port will compile failed with datetime.cpp(26): error C2039: 'runtime_error': is not a member of 'std'

### DIFF
--- a/source/datetime.cpp
+++ b/source/datetime.cpp
@@ -15,7 +15,7 @@
 	#include <windows.h>
 #endif
 #include <array>
-#include <ostream>
+#include <stdexcept>
 
 namespace {
 	std::tm localtime()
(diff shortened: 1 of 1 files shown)


SteffenL:
If `std::runtime_error` is missing because of a missing include, wouldn't it be better to include `stdexcept` instead of `ostream`?

LilyWangL:
> If only `std::runtime_error` is missing because of a missing include, how about including `stdexcept` instead of `ostream`?

@SteffenL Done.

SteffenL:
Although I haven't tested this or come across the problem you were facing, it looks good to me.

Unfortunately, as much as I would like to accept improvements, I have no power to do so.

cnjinhao:
Thanks for your PR. This is a patch I have checked it into the `hotfix-1.7.3` branch.

Issue #524, opened 2020-03-21
Label documentation says url shouldn't work


KingDuckZ:
[Current documentation page](http://nanapro.org/en-us/documentation/widgets/label.htm) for `nana::label` says:

> Arrow cursor becomes a hand when the cursor moves over the red text. Meanwhile, it calls a web browser to open the URL by clicking the red text. NOTE: the url only works under Windows

with the *NOTE* part in bright red. I found this is not true as opening urls works just fine for me on KDE Gentoo Linux.

cnjinhao:
What if no `firefox in the` system. See the `posix_open_url` function

https://github.com/cnjinhao/nana/blob/6156acf068d392b9d19fca7ec685443fa9dd1dcc/source/system/platform.cpp#L31

KingDuckZ:
I see, so it's a coincidence since I'm running firefox myself!

Looking a bit into it, it seems the best way is to invoke `xdg-open`, which I can confirm was already on my system. It's a bash script and at first look it seems to contain logic to detect the desktop in use and launch the user's default browser based on that. So maybe before/instead of checking if firefox exists, you should check if xdg-open exists and eventualy run that?

KingDuckZ:
I just made a PR, it works for me (TM). I can only test it on Gentoo unfortunately. However I think users are more likely to have xdg-open on their system than firefox, so I'm hoping it works more often than the previous solution.

PR #527, opened 2020-03-21
Develop xdgopen

KingDuckZ:
Use xdg-open like I suggested on issue #524


cnjinhao:
Thank you a lot for this PR.

PR #531, opened 2020-03-22
Initialise everything to make sure accel_key is constexpr.

KingDuckZ:
Need to initialise everything to make sure classes are usable in a constexpr context.
This is useful for my helper `make_accel_key()` function which I will share shortly.


cnjinhao:
Thank you a lot for this PR

Issue #525, opened 2020-03-21
Question about menubar shortcuts


KingDuckZ:
I looked everywhere in the documentation but I can't find anything about this.
I want to add common shortcuts to my menu entries, for example ctrl+q should do the same as "File -> Exit". Ideally, it should be possible for end-users to customise this, but this is not currently very important.
How can I add a shortcut in nana?

ErrorFlynn:
The library supports keyboard accelerators, but that's not documented at the moment. Take a look at this old forum thread: http://nanapro.org/en-us/forum/index.php?u=/topic/335/ggdoes-the-library-provide-a-way-to-implement-hotkeys

KingDuckZ:
Thanks @ErrorFlynn, it works for me on Gentoo :) I'm trying to improve the setup of shortcuts a little bit, see my PR #532 for more details. Also, unfortunately this doesn't show the shortcut keys next to the menu entry. I guess there's no way to do that yet?

ErrorFlynn:
>Also, unfortunately this doesn't show the shortcut keys next to the menu entry. I guess there's no way to do that yet?

I'm not sure if that should be done by the library, it's not so hard to do it yourself. For example:

````
#include <nana/gui.hpp>
#include <nana/gui/widgets/menubar.hpp>

int main()
{
	nana::form fm;
	nana::menubar mb {fm};
	auto cb_exit = []
	{
		nana::API::exit();
	};
	mb.push_back("&File").append("E&xit\t\t\t\tCtrl+Q", 
		[&](nana::menu::item_proxy&) { cb_exit(); });
	nana::accel_key acc_quit;
	acc_quit.key = 'q';
	acc_quit.ctrl = true;
	fm.keyboard_accelerator(acc_quit, cb_exit);
	fm.show();
	nana::exec();
}
````

PR #532, opened 2020-03-22
Allow invoking a menu action by index.

KingDuckZ:
@cnjinhao I think I need your input on this. Following the hint I was given on issue #525 I tried to add shortcuts to my menu but that created lots of code duplication. For example:

```cpp
my_menubar.at(0).append("E&xit", [](nana::menu::item_proxy&){nana::API::exit();});
my_form.keyboard_accelerator(accel, [](){nana::API::exit();});
```

Since the functor signature is different it's quite hard to re-use the same functor. So now if I want to show an "Are you sure?" dialog I have to remember to do it in two places and create clutter. I'm sure we agree on the problem here.

My idea then was to be able to do this:

```cpp
template <typename F>
void append_to_menu (nana::form& owner, nana::menu& menu, const std::string& text, const F& func, nana::accel_key accel) {
    menu.append(text, func);
    owner.keyboard_accelerator(accel, [&menu,idx=menu.size()-1](){menu.invoke(idx);});
}
```

but I obviously need to be able to trigger the menu entry functor from outside. My code above works form my simple case (with the changes from this PR) but I'm not too sure about my `invoke()` implementation. For example I'm not sure it will work for submenus. What are your thoughts? Was there such a feature already? If not, how can I improve my PR?

### DIFF
--- a/include/nana/gui/widgets/menu.hpp
+++ b/include/nana/gui/widgets/menu.hpp
@@ -171,6 +171,7 @@ namespace nana
 		std::size_t size() const;					///< Return the number of items.
 		int send_shortkey(wchar_t key);
 		void pick();
+		int invoke(std::size_t pos);				//< Send click event to the given entry.
 
 		menu& max_pixels(unsigned);				    ///< Sets the max width in pixels of the item.
 		unsigned max_pixels() const;
--- a/source/gui/widgets/menu.cpp
+++ b/source/gui/widgets/menu.cpp
@@ -1342,6 +1342,28 @@ namespace nana
 				impl_->window_ptr->pick();
 		}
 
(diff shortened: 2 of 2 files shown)


KingDuckZ:
On a second thought I think I realise this is a hackish solution.
`menu::append()` should really accept an `accel_key`, but I don't know how to retrieve the top-level form from a menu object to register the accel key. Also, I think `nana::form` code is missing a remove method, because if you remove an entry from a menu then the corresponding shortcut should also be removed from the form.

cnjinhao:
>>menu::append() should really accept an accel_key

I agree on this approach. I think we should implement it in version 1.8

PR #537, opened 2020-04-11
Use iswalnum in text editor

call-me-void:
When you use double-click selection in text editor with non-ascii text, you'll get this assertion

assert

simple fix is to use iswalnum, alongside with already used iswspace.


cnjinhao:
Thank you a lot for this PR

PR #543, opened 2020-05-20
modified toolbar disabled icons drawing

besh81:
Ciao,

I modified the way toolbar draw the icons of disabled items.
What I didn't like was the gray square around the disabled icons even if the icons had transparent pixels.
Basically I modified the graphics::rgb_to_wb function adding a flag to skip the transparent pixels of the icons.
Here the result:
toolbar

For sure there are better way to implement this behavior... I chose the easier without digging to deep into graphic/image code.


cnjinhao:
Thanks for your PR!

PR #544, opened 2020-05-22
Debugging place

qPCR4vir:
I have some problems in windows 7. 
v1.7.4 is OK, but in 1.8, commit 028ad17a11274a0f0db73541766da17daad3995b from 11.03.2020 15:52:34	"the place weight integer value is DPI-dependent" introduced an infinite loop in collocate from group. 
The loop is in place.cpp:
`while ((rest_px > 0) && blocks)`
in function `_m_revise_adjustable`
Commit 70e096210fad075449353b372af594ea2475f3ed seems to fix that, but not sure if not introduces another bug. Anyway, most uses of place are broken, even if no group is used.
The error in place is still there. 
I'm trying to follow some ideas: place relied in double comparisons for equality that used to work, but after introduction to dependence on DPI don't work anymore. 
Also in source\gui\detail\bedrock_windows.cpp there are variables not used: auto dpi_x = HIWORD(wParam); 	auto dpi_y = LOWORD(wParam);

This PR

-  reintroduces a sln and project for VS2019,
-  actualizes the select_filesystem.cmake, 
- fix compilation error in animation
- avoid breaking compatibility in constructor of folderbox
- etc (code comments, warnings fixes).

There is some debug code to try to fix place that need to be eliminate by searching for std::cout in place.cpp

### DIFF
--- a/build/cmake/select_filesystem.cmake
+++ b/build/cmake/select_filesystem.cmake
@@ -37,14 +37,14 @@ if (CXXBuiltIn_FS_BuiltIn)
     message (STATUS "C++ Filesystem library:     builtin")
 
 # or we need stdc++fs library ?
-else()
+else(CXXBuiltIn_FS_BuiltIn)
     set (CMAKE_REQUIRED_LIBRARIES "${CMAKE_REQUIRED_LIBRARIES_ORIGINAL} stdc++fs")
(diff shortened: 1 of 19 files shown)


cnjinhao:
Thanks a lot for your PR!

PR #547, opened 2020-05-29
Hotfix 1.7.4 new toolbar interface

besh81:
Hi,
I worked a little on the toolbar here a brief list of changes:
- redesigned toolbar interface
- added dropdown tool
dd

More in detail:
1) added a new version of the append member functions where the first parameter is the tool-type and the last (optional) the event handler. The event is raised when the tool is clicked. The old append member functions are marked as deprecated.

2) added tools_height. This is a replacement for the scale function marked as deprecated. The new name seems "more clear" to me.

3) added append_separator. This is a replacement for the separate function marked as deprecated. Same as before.

4) added count to gets the number of tools and separators.

5) added at(index) to access the tool at specified position.

6) marked as deprecated some of the member functions I previously added. Now should be accessed only from the tool item_proxy. No reason to have these function directly on the toolbar interface.

7) added dropdown tool (see image above). Added the following item_proxy functions:
   - dropdown_append: adds item to the dropdown menu
   - dropdown_answerer: sets the event handler for the dropmenu item

For sure I'm forgetting some change I made...
... anyway let me know what you think.

Ciao

### DIFF
--- a/include/nana/gui/widgets/float_listbox.hpp
+++ b/include/nana/gui/widgets/float_listbox.hpp
@@ -54,7 +54,8 @@ namespace nana
 
 				virtual ~item_renderer() = default;
 				virtual void image(bool enabled, unsigned pixels) = 0;
-				virtual void render(widget_reference, graph_reference, const nana::rectangle&, const item_interface*, state_t) = 0;
+				virtual void background(widget_reference, graph_reference) = 0;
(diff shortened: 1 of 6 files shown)


cnjinhao:
Thanks a lot for your PR! Very useful feature, well done!

PR #550, opened 2020-06-11
apply Webkit coding style

mikeandmore:
Nana's current coding style is very inconsistent. This patch uses clang-format to format the entire code base.

I choose to apply Webkit's coding style because it's also indenting using 4 spaces. We can discuss what works better.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,137 @@
+---
+Language:        Cpp
+# BasedOnStyle:  WebKit
+AccessModifierOffset: -4
+AlignAfterOpenBracket: DontAlign
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   false
+AlignTrailingComments: false
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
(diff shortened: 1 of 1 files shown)


qPCR4vir:
Sorry, this is a massive change in source code with make very difficult to trace real evolution of code change. Please dont merge it.
It will be better as just an "issue" proposing some code style for discussion. You can also add just the .clang-format file to be used as starting point for experimentation. @cnjinhao could use it to reformat a few files a time, review the results and commit only a few change each time. 
I personally agree that some reformatting is need, but not automatically in full. In many places there are manually crafted formatting that make the code much more readable.

mikeandmore:
> Sorry, this is a massive change in source code with make very difficult to trace real evolution of code change. Please dont merge it.
> It will be better as just an "issue" proposing some code style for discussion. You can also add just the .clang-format file to be used as starting point for experimentation. @cnjinhao could use it to reformat a few files a time, review the results and commit only a few change each time.

Yeah, I guess it depends on how do you make the trade-offs. You can pay an up-front cost of reformatting, or scatter them into many small changes. I've pushed just the `.clang-format`, feel free to apply.

> I personally agree that some reformatting is need, but not automatically in full. In many places there are manually crafted formatting that make the code much more readable.

True, but I think majority of reformatting works well with `.clang-format`. Perhaps most importantly, this is can be a guideline for other contributors.

PR #554, opened 2020-06-25
Fix : Cmake build error on macOS

wwriter:
I tried building nana library on my mac using cmake.
While I was working on it, the output said some necessary header was not included.
So I included the header file and now I can get libnana.a


cnjinhao:
Thanks for your PR!

Issue #563, opened 2020-09-09
place and toolbar issue


besh81:
Hi,

I found a bug using place and toolbar: 

- when the toolbar is inside a vertical field (not covering all the available width) and the window is resized the toolbar become 100% width covering the adjacent fields

Tested on master, hotfix and develop branches.

Sample code:
```
#include <nana/gui.hpp>
#include <nana/gui/place.hpp>
#include <nana/gui/widgets/textbox.hpp>
#include <nana/gui/widgets/toolbar.hpp>

int main()
{
	nana::form fm(0, nana::size{ 300, 200 });

	nana::place pl{ fm };
	pl.div("<<vert weight=100 aaa><>>");

	nana::textbox t{ fm };
	pl["aaa"] << t;

	nana::toolbar tb{ fm };
	pl["aaa"] << tb;
	tb.bgcolor(nana::color(255, 0, 0));

	pl.collocate();
	fm.show();
	nana::exec();
}
```

![Cattura](https://user-images.githubusercontent.com/16460595/92567300-c816e980-f27d-11ea-8014-5f43904267f6.JPG)

cnjinhao:
The toolbar is automatically resized when parent is resized.
https://github.com/cnjinhao/nana/blob/master/source/gui/widgets/toolbar.cpp#L319-L324

besh81:
Ok, I removed the `event_size` event from the `attached` function and everything seems to work better.
I also check the other widgets and no trace of the `event_size` event inside the `attached` function.
I didn't notice this behavior before because I usually use the toolbar at the top of the window with the full width (as everybody does).

My proposal is to remove the `event_size` event from the `attached` function.

Thanks

PR #568, opened 2020-09-24
Removed resize event from toolbar

besh81:
Removed resize event from toolbar to avoid the control is automatically resized when parent is resized   (issue #563)


cnjinhao:
Hi, if we remove the resize handler, the codebase that uses toolbar without using place will cause bugs. I think we should add a new method to remove the resized event handler, that could prevent breaking existing codebase.

besh81:
I consider the toolbar a control like any other. If apps (without place) needs to adjust controls size when the parent size change they should add the proper resize event. I don't think the lib should add events for specific cases.

cnjinhao:
Agree with you, it's a legacy problem.

PR #572, opened 2020-10-24
added a feature : registering multiple tray icon animations

wwriter:
I thought it would be nicer if a user can register multiple tray icon animations.
At first, I tried to give `period` function a new name such as `start_animation`.
Then I realized this is a minor version update, which has to ensure the backward compatibility.
So I modified some of the codes and added some new functions in a way that it won't cause a side effect.


cnjinhao:
Thanks alot for your PR!

PR #573, opened 2020-10-30
fix combox rendering issues

ErrorFlynn:
This commit solves three combox rendering issues:

1. if one or more list items has an image, and the combox is editable, the background of the image area to the left of the text editor isn't drawn (resulting in a black background area to the left of the text)
2. if one or more list items has an image, the image is not centered vertically when drawn to the left of the text editor
3. if the combox is not editable, the background of text editor area uses predefined constants for the color gradient, ignoring the widget's scheme bgcolor

combox_bugs

### DIFF
--- a/source/gui/widgets/combox.cpp
+++ b/source/gui/widgets/combox.cpp
@@ -201,33 +201,33 @@ namespace nana
 					{
 						editor_->editable(enb, false);
 						editor_->show_caret(enb);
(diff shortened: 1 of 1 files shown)


cnjinhao:
Thanks a lot for your PR!

cnjinhao:
Applied this change to `develop-1.8`

PR #575, opened 2020-11-04
renamed some variables to prevent shadowing warning

crillion:
Hi, 

   I normally compile with g++ using the warning switch -Wshadow=local , which gives me an alert when a variable in a function is shadowing another one such as a parameter of the function itself. By this switch I've noticed that a couple of files in nana have this problem (combox.hpp and place.hpp), so I've changed there the name of the variable.
   
    Recompiled nana and tested : I get no more shadow warnings in my programs and they work fine.
	
	
   Thanks,
   
         Marco


cnjinhao:
Hi, @crillion , this PR has been merged into branch `develop-1.8`. Thanks for your PR!

Issue #585, opened 2020-12-14
Building on VS 2019 results in syntax error


khrynczenko:
Hi,
Today I tried build nana using `v1.7.4` tag, unfortunately with no success. There is a syntax error that results in a build failure. The fail looks as follows.
```
[build] ..\source\system\split_string.cpp(21): error C2146: syntax error: missing ')' before identifier 'and'
[build] ..\source\system\split_string.cpp(21): error C2065: 'and': undeclared identifier
[build] ..\source\system\split_string.cpp(21): error C2146: syntax error: missing ';' before identifier 'sep'
[build] ..\source\system\split_string.cpp(21): error C2059: syntax error: ')'
[build] ..\source\system\split_string.cpp(22): error C2059: syntax error: ';'
[build] ..\source\system\split_string.cpp(25): error C2065: 'start': undeclared identifier
[build] ..\source\system\split_string.cpp(25): error C2065: 'end': undeclared identifier
[build] ..\source\system\split_string.cpp(25): error C2065: 'start': undeclared identifier
[build] ..\source\system\split_string.cpp(28): error C2059: syntax error: 'return'
[build] ..\source\system\split_string.cpp(32): error C2059: syntax error: '}'
[build] ..\source\system\split_string.cpp(32): error C2143: syntax error: missing ';' before '}
```

For some reason inside the `split_string.cpp` file,  there is an `and` keyword used, which causes problems on VS 2019. The possible solution is to either `#include <ciso646>` or just replace it with `&&`. Personally, I would do the latter.

Code that causes the problem:
https://github.com/cnjinhao/nana/blob/57f918f00d9b21a2a11655eabd49665b9b4dd760/source/system/split_string.cpp#L21

Edit:
FYI: I am using CMake (3.18.5) together with the newest CMakeTools and VS Code, and Visual Studio 2019 compiler (16.2.2).

khrynczenko:
Fixed in #586

PR #586, opened 2020-12-14
Fix syntax error when building with VS 2019

khrynczenko:
Hi,
This is a quick fix for #585. It removes the syntax error, when building using Visual Studio, by replacing `and` keyword with `&&` in one of the conditions.


cnjinhao:
Thank you for fixing this bug.

PR #587, opened 2020-12-16
fix bug in toolbar dropdown items where disabled items were highlighed on mouse hover

besh81:


besh81:
Add another fix regarding menu

cnjinhao:
Thanks for your fix!

PR #596, opened 2021-01-07
Sligth optimization and refactor code

GermanAizek:
Compiler handles code better with more localization of variables, code-style fixes and memory leak fixes.

### DIFF
--- a/include/nana/gui/compact.hpp
+++ b/include/nana/gui/compact.hpp
@@ -38,8 +38,10 @@ namespace nana
                 auto p = new Form(std::forward<Args>(args)...);
 
                 if (p->empty())
+                {
+                    delete[] p;
                     throw std::runtime_error("form_loader failed to create the form");
-
+                }
 
                 detail::form_loader_private::insert_form(p);
                 if (IsVisible)
--- a/include/nana/gui/widgets/skeletons/textbase.hpp
+++ b/include/nana/gui/widgets/skeletons/textbase.hpp
@@ -253,7 +253,6 @@ namespace skeletons
 
 				if (is_unicode)
 				{
-					const char * le_boms[] = { "\xEF\xBB\xBF", "\xFF\xFE", "\xFF\xFE\x0\x0" };	//BOM for little-endian
 					int bytes = 0;
 					switch (encoding)
 					{
@@ -266,6 +265,7 @@ namespace skeletons
 					}
 
 					if (bytes)
+						const char * le_boms[] = { "\xEF\xBB\xBF", "\xFF\xFE", "\xFF\xFE\x0\x0" }; //BOM for little-endian
 						ofs.write(le_boms[static_cast<int>(encoding)], bytes);
 
 					for (std::size_t pos = 0; pos < count; ++pos)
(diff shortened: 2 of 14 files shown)


include/nana/gui/widgets/skeletons/textbase.hpp
```diff
@@ -266,6 +265,7 @@ namespace skeletons
 					}
 
 					if (bytes)
+						const char * le_boms[] = { "\xEF\xBB\xBF", "\xFF\xFE", "\xFF\xFE\x0\x0" }; //BOM for little-endian
```
beru, line 268: I'm afraid this modification will change original flow of the code since there aren't curly brances after `if` statement.
So it should be like this.
```cpp
				if (bytes) {
						const char * le_boms[] = { "\xEF\xBB\xBF", "\xFF\xFE", "\xFF\xFE\x0\x0" }; //BOM for little-endian
						ofs.write(le_boms[static_cast<int>(encoding)], bytes);
				}
```

And I'm not really sure this subtle change would actually gives us better optimized code.
Have you actually checked assembly code and measured performance difference before and after?

GermanAizek: @beru,
Thank you for reviewing the code, any localization of variables simplifies the work of the compiler. The increase is small, of course, but these edits do not particularly spoil the style of the code.

include/nana/internationalization.hpp
```diff
@@ -139,7 +139,7 @@ namespace nana
 		template<typename Return>
 		void _m_add_args(std::function<Return()> fn)
 		{
-			args_.emplace_back(new arg_function<Return>(fn));
+			args_.push_back(std::unique_ptr<arg_function>(new arg_function<Return>(fn)));
```
qPCR4vir, line 142: Hi, could you explain why is this better?

include/nana/gui/compact.hpp
```diff
@@ -38,8 +38,10 @@ namespace nana
                 auto p = new Form(std::forward<Args>(args)...);
 
                 if (p->empty())
+                {
+                    delete[] p;
                     throw std::runtime_error("form_loader failed to create the form");
-
+                }
```
qPCR4vir, line 44: Why not delete p?
... hmm, and why `delete[] p` and not just `delete p`?

include/nana/gui/widgets/skeletons/textbase.hpp
qPCR4vir: hmm.. did you compile this?
`le_boms `is not defined at the moment it is "used" in the next line outside the if were it is defined.

qPCR4vir:
Hi, @GermanAizek thank you for your effort. But I'm sorry to say: before you open a PR, please compile AND run the tests.
See here:
https://travis-ci.org/github/cnjinhao/nana

source/gui/place.cpp
```diff
@@ -862,14 +862,13 @@ namespace nana
...
 				{
-					auto child_floor = child->calc_weight_floor(wd);
-
 					if (child->is_percent())
 					{
 						ratio += child->weight.real();
 					}
 					else
 					{
+						auto child_floor = child->calc_weight_floor(wd);
```
cnjinhao, line 871: This modification will cause malfunctions. Think about this situation.
`<><weight=60% <a><b fit>>` It doesn't adjust the widgets in `field b` to fit their contents.

qPCR4vir (changes requested):

PR #598, opened 2021-01-14
fix bug in menu where item_proxy index is initialized with the wrong value

besh81:


cnjinhao:
Thank you for your PR!

PR #607, opened 2021-02-04
fix checkbox square misaligned with text

ErrorFlynn:
before:
cbox_before

after: 
cbox_after


cnjinhao:

PR #609, opened 2021-02-07
Introduce command to share with menu and toolbar

qPCR4vir:
Hi, 
The goal is to: 
## Introduce command to share with menu, toolbar, button and checkbox

please, compare: [widget_show2 manual item creation](https://github.com/qPCR4vir/nana-demo/blob/develop/widget_show2.cpp#L317)
with: [widget_show2 created with command](https://github.com/qPCR4vir/nana-demo/blob/command/widget_show2.cpp#L324)

you will need to review and redesign a few thing, but I hope this is the first draft of what we wanted.

- `menu` vs. `toolbar`:
toolbar is a widget, menu - not. 
   -  `menu`: hold a plain ptr to an `implement` which use structures in a `drawerbase::menu` namespace. 
The later includes a `menu_type`, `menu_item_type`, and the enum `checks`, 
```
			enum class checks
			{
				none,
				option,
				highlight
			};
```
`menu_item_type` hold a plain pointer to a `menu_type` and all the state-data of the item which can be acceded with an `item_proxy` which hold a pointer to the `menu `and the position order of the item.
 A new item can be added with `menu::append` or `menu::insert`.
```
			enum class tools
			{
				separator,
				button,
				toggle,
				dropdown
			};
```
  - `toolbar`: the `drawer `hold a plain pointer to a `drawer_impl_type` that hold an `item_container` of `toolbar_item`s. The later hold the state-data. 
  A new item can be added with `toolbar::append` or with `item_proxy::dropdown_append`.
- [ ] Implement/use `checks`/`tools`
- [ ] create `command`: a no-widget class, wich state-data useful to set and to share `menu`, `toolbar`, `button `and `checkbox`
  - [x] title/text, tooltip/status
  - [x] event handler 
  - [ ] `image` (a shared type),  
  - [ ]  set/get `image` 
  - [ ] checks style
  - [x] flags: enabled, checked

- [ ] Make menu and toolbar API similar:
  - [ ] prefer item_proxy

In `menu`:
- [ ] no struct `arg_menu` ? 
- [x] menu::append(shared_command) 
- [x] menu::insert(shared_command) .
- [ ] in `menu_type`
- [ ] in `menu_item_type` <---- real data here (generalize?)
    - [x] hold a shared_command
    - [x] add constructor that take a `shared_command`
    - [ ] set/get `shared_command`  (to be used by other constructors)
    - [ ] really need 'linked'?
- [ ] add to item_proxy 
  - [ ] set/get `shared_command`  (to be used by other constructors)
  - [ ] set/get `image` 
  - [ ] answerer()
- [ ] reuse renderer_interface?
In toolbar:
- [ ] `dropdown_item` construct take a `menu` and vice ferse?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,8 @@ before_install:
   # donwload nana-demo first
   # we are in:  'user'/nana/
   - cd ..
-  # we are in:   'user'/
-  - git clone --depth=1 --branch=master https://github.com/qPCR4vir/nana-demo.git nana-demo
+  # we are in:   'user'/   ------->>  todo: set branch back to develop or master
+  - git clone --depth=1 --branch=command https://github.com/qPCR4vir/nana-demo.git nana-demo
   # now we have 'user'/nana-demo, 'user'/nana/  and we are in: 'user'/
   - export PATH="$HOME/bin:$PATH"
   - wget --no-check-certificate --no-clobber -O /tmp/tools/cmake https://cmake.org/files/v3.12/cmake-3.12.0-rc3-Linux-x86_64.sh || true
--- a/build/vc2019/nana.vcxproj
+++ b/build/vc2019/nana.vcxproj
@@ -253,6 +253,7 @@
     <ClInclude Include="..\..\include\nana\gui.hpp" />
     <ClInclude Include="..\..\include\nana\gui\animation.hpp" />
     <ClInclude Include="..\..\include\nana\gui\basis.hpp" />
+    <ClInclude Include="..\..\include\nana\gui\command.hpp" />
     <ClInclude Include="..\..\include\nana\gui\dragdrop.hpp" />
     <ClInclude Include="..\..\include\nana\gui\dragger.hpp" />
     <ClInclude Include="..\..\include\nana\gui\drawing.hpp" />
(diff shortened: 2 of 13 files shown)


cnjinhao:
Very good idea!

>no struct arg_menu ?

Because `menu::item_proxy` is like a handle/proxy of a menu item, not an `event_arg`.

>really need 'linked'?

What does `linked` mean? Does it mean that `dropdown_item` directly refers to the shared `command`?
I think it would be better to make `dropdown_item` directly refers to the shared `command`, in default, and make it Copy-on-Write. Then the redundant data members can be eliminated.

>reuse renderer_interface?

Yes, the renderer is cloneable. We can make tiny changes based on the current renderer just with a small piece of code.

>dropdown_item construct take a menu and vice ferse?

Sorry, I don't understand, would you please explain it.

qPCR4vir:
> > really need 'linked'?

> What does `linked` mean? 

oh, sorry, I just wanted to know how `menu_item_type::linked` member works.

> Does it mean that `dropdown_item` directly refers to the shared `command`?
> I think it would be better to make `dropdown_item` directly refers to the shared `command`, in default, and make it Copy-on-Write. Then the redundant data members can be eliminated.

yes, something like that is a good solution! It is not trivial to have a really good solution. I only implemented the simplest possibility, and then, maybe, you will have some day time to make it better! I will try to implement a few more details, but for a robust design one need more experience.

> > dropdown_item construct take a menu and vice verse?
> 
> Sorry, I don't understand, would you please explain it.

My bad. I mean that a sub-`menu` and a `toolbar` dropdown are both a kind of containers of "command-items" and it could be great if we could just pass a menu to toolbar::append or item_proxy::dropdown_append to add all the items in one instruction.

PR #610, opened 2021-02-08
unify checkbox look

StillGreen-san:
- replaced radio bitmaps with calls to round_rectangle
- changed menu check mark to be same as in checkbox

Before - After (on Windows)
2021-02-08_14-18-48_nnp-test1 2021-02-08_14-18-57_checkbox


qPCR4vir:
why?

StillGreen-san:
For consistency and easier theming.
And while it's probably possible to change the bitmaps to be affected by the color settings, I thought it was cleaner to have them generated in code.

qPCR4vir:
yes, I agree, ... but is just not very pretty. Maybe there is a way to make them more beautiful? scalable? optional?
sorry

StillGreen-san:
If we are talking about the radio buttons (top right) then they could be made smaller, which also makes them rounder. But I'm not sure if that's better.
2021-02-08_19-52-39_checkbox

ErrorFlynn:
In my opinion, the benefit of the proposed unified look is not worth it, because the new graphics are just too ugly in their current form (especially the radio buttons).

That being said, I do have a problem with the checkbox using bitmaps: the library always renders them at 16x16 pixels, regardless of the height of the text. That can be an issue when the desktop uses a high scaling factor, especially if the font size for the widget is larger than the default size. I might try to do something about that in the near future (I'm thinking to first try to draw them in an acceptable way somehow, and if that fails, stretching the bitmaps might be an option).

PR #613, opened 2021-03-12
add listbox::items() method to get all items index_pairs

riquems:
Hi, I had problems into getting all items from a listbox, so I came with this idea of contributing to the project. It's a very small change but I think it's very useful :).

If the functionality already exists let me know, I found only one way to reproduce what I wanted:
Select the category items and then get the selected items (which wasn't very intuitive at first and it works only for individual categories), like:
```c++
lb_listbox1->at(0).select(true);
listbox::index_pairs items = lb_listbox1->selected();
```

Now I can do this:
```c++
listbox::index_pairs items = lb_listbox1->items();
```

Anyway, It worked in my tests, feel free to edit/implement in your way, I just wanted to see this feature in the library. Thanks for your awesome job.

### DIFF
--- a/include/nana/gui/widgets/listbox.hpp
+++ b/include/nana/gui/widgets/listbox.hpp
@@ -1590,6 +1590,7 @@ the nana::detail::basic_window member pointer scheme
 		///< Prevent sorting until `freeze` is set to false.
 		bool freeze_sort(bool freeze);
 
+		index_pairs items() const;		    ///<Get the absolute indexes of all items
 		index_pairs selected() const;		///<Get the absolute indexes of all the selected items
 
 		void show_header(bool);
--- a/source/gui/widgets/listbox.cpp
+++ b/source/gui/widgets/listbox.cpp
@@ -1684,6 +1684,25 @@ namespace nana
 				return changed;
 			}
 
+			index_pairs pick_items() const
+			{
+				index_pairs results;
+				index_pair id;
+
+				for (auto& cat : categories_)
+				{
+					id.item = 0;
+					for (auto& m : cat.items)
+					{
+						results.push_back(id);  // absolute positions, no relative to display
+
+						++id.item;
+					}
+					++id.cat;
+				}
+				return results;
+			}
+
 			/// return absolute positions, no relative to display
 			/**
 			 * @param for_selection Indicates whether the selected items or checked items to be returned.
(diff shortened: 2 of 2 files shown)


cnjinhao:
I just wonder why we need this method. IMO, it looks like we get all indexes of an array.

riquems:
> 
> 
> I just wonder why we need this method. IMO, it looks like we get all indexes of an array.

I mean, it's self explanatory actually: get all items from the listbox. If there's another cleaner way to do this... I don't know.

If you can get the selected items from the listbox, why can't you get all of them? :)

qPCR4vir:
Hi, @riquems ...
I will try to come back... not sure yet.
How are you going to use the returned list of all indexes?
It is not he same to try to get the items the user selected and you need to do something with that (there is no other way to do that) than to get "all the items". "All the items" are always there, there is no particular selection from the user. 
The user make a selection of items and then click something and expect the program uses that selection: you "stop", get the indexes, and do that.
If you want to use all the items you don't need to ask: the list has all of then. Maybe, there could be a better way to visit each of then. But to keep a list of all indexes don't seems to be very good idea. If you have an array you don't need a list of the indexes to work with the items in the array. If somebody have a selection of items in the array, well, you will need to get a list of the indexes of the seleted items. 
To visit each item you could:
```
for (int c=0; c<listb.size_categ(); c++)
  for (int i=0; i<listb.size_item(c); i++)
    listb[c][i]; // the items
```
I agree, it could be better (to be implemented):
```
for (auto &cat : listb)
  for (auto &i : cat)
    i;  //  the items
```

riquems:
Hi @qPCR4vir, thanks for coming back! This is my use case:

image

I have two listboxes, and I can move items from one to the other. The simple arrow head (> and <) moves only selected items, and the dual arrow heads (>> and <<) moves all items from one list to the other. To do this, I figured it out by using index_pairs and basically moving the items from one list to the other (welp, thats not so descriptive, here is my demo code: https://pastebin.com/ckG8jbDD)

Without the method I implemented, the code that moves all items looks like:

```
lb1.at(0).select(true);
auto items = lb1.selected();

move_items(items, lb1, lb2);
```

This can be found in line 29 of the pastebin code.

With the method I implemented, you can do this:

```
auto items = lb1.all();

move_items(items, lb1, lb2);
```

Which is more readable.

I didn't know we could iterate the listbox items like you did with the for each thing. That's cool, but imperative. A method to return these in a way we can populate another listbox with the same items, is still better.

Welp, maybe my code isn't ideal either, that's the way I did it though. I'm open for suggestions for improving my code (if I went the hard path), or suggest another possible better way of doing it. I just went conservative and followed what has been already done in the .selected() method.

PR #614, opened 2021-03-25
Now double click in spinbox select all the number (or text)

besh81:
Changed behavior of double-click event in spinbox control.
BEFORE: double-click was the same of click
NOW: double-click select all the number (or text) inside the spinbox


cnjinhao:
Thanks for your PR!

PR #615, opened 2021-04-08
Fixed a problem that borderless was not applied to buttons

JannLee:
test code
```cpp
btn.transparent(true).edge_effects(false).enable_focus_color(false).borderless(true);
btn.transparent(true).edge_effects(false).enable_focus_color(false).borderless(false);
```


cnjinhao:
Thanks for your PR!

PR #626, opened 2021-07-07
added a feature : disabling tabstop

wwriter:
There are widgets that always use tabstop feature (such as `button`, `textbox`, `spinbox`), because it is set at their creation time.
I think it is likely that someone might want to disable the tabstop feature for some widgets.
I got the idea from `eat_tabstop` api and changed `tabstop` function somewhat similar to that.

If nana library has already provided an api that does the same thing, please let me know.
(I'll use that instead and close this pull request)


cnjinhao:
Thanks for your PR!

PR #645, opened 2022-01-27
listbox scheme additions (item border & category)

ErrorFlynn:
Currently, the color of the item selection border, and the category foreground color are hard-coded, limiting the ways in which the listbox appearance can be changed (for example, a dark background can't be used). This PR solves the issue by making those colors modifiable by the user.


cnjinhao:
Thanks for your PR

PR #651, opened 2022-02-03
Fix warnings in clang

linquize:
Remove std::move from temporary variables
Do not capture unused variable


cnjinhao:
Thanks for your PR!

PR #653, opened 2022-02-23
convert xk_home xk_end correctly

rhos:
Home and End keys don't work on linux, they do nothing.
Seems like a correct fix


cnjinhao:
Thank you for your PR!

PR #665, opened 2022-06-29
Fixed int conversions in pixel_buffer::ellipse.

ventsin:
Fixes #664.

@cnjinhao please merge this as this branch cannot be built under MSVC right now.

Issue #664, opened 2022-06-29
pixel_buffer::ellipse can't be compiled in MSVC


ventsin:
The method `void pixel_buffer::ellipse(const nana::point& pt, float a, float b, const color& c)` was added in commit bf26f622e382383969ce00c729b7ee30f6671f8b, and it breaks compilation under MSVC.

Compile log as follows:
```
[...]source\paint\pixel_buffer.cpp(1564,15): warning C4244: 'initializing': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1565,15): warning C4244: 'initializing': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1568,18): error C2398: Element '2': conversion from 'float' to 'int' requires a narrowing conversion 
[...]source\paint\pixel_buffer.cpp(1568,18): warning C4244: 'argument': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1569,18): error C2398: Element '2': conversion from 'float' to 'int' requires a narrowing conversion 
[...]source\paint\pixel_buffer.cpp(1569,18): warning C4244: 'argument': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1572,12): warning C4244: '=': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1574,20): warning C4244: '=': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1599,15): error C2398: Element '1': conversion from 'float' to 'int' requires a narrowing conversion 
[...]source\paint\pixel_buffer.cpp(1599,15): warning C4244: 'argument': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1600,15): error C2398: Element '1': conversion from 'float' to 'int' requires a narrowing conversion 
[...]source\paint\pixel_buffer.cpp(1600,15): warning C4244: 'argument': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1602,12): warning C4244: '=': conversion from 'float' to 'int', possible loss of data 
[...]source\paint\pixel_buffer.cpp(1604,20): warning C4244: '=': conversion from 'float' to 'int', possible loss of data
```

Basically, `a` and `b` need to be rounded before being added to `x` and `y`, respectively.

PR #666, opened 2022-07-01
Fix compilation error and run-time error with VS2022-17.1 (v143) and c++std20

qPCR4vir:
Hi,
please, don't take too serious this - you may have better solutions.

The first is a variation on #664 and #665 becouse of compilation error. You may want to use std::round or simmilar if needed.

The second is a fix for a run-error (crash) that invalidate the` .end()` after an` .erase()`. I don't know why, but it appear in `release `and `debug `mode. The debuger showed undefined `end()`, which coincide with the documentation. I used trivial index. Not very elegant but works.

### DIFF
--- a/source/gui/detail/window_manager.cpp
+++ b/source/gui/detail/window_manager.cpp
@@ -1353,19 +1353,19 @@ namespace detail
 			internal_scope_guard lock;
 
 			auto& safe_place = impl_->safe_place.table();
-			for (auto i = safe_place.begin(); i != safe_place.end();)
+			int i=0;
+            while (i < safe_place.size()) 
 			{
-				if (i->first->thread_id == thread_id)
+                if (safe_place[i].first->thread_id == thread_id)
 				{
-					auto functions = std::move(i->second);
-					i = safe_place.erase(i);
+					auto functions = std::move(safe_place[i].second);
+                    safe_place.erase(safe_place.begin()+i);
 
 					for (auto& fn : functions)
 						fn();
-				}
-				else
-					++i;
-			}
+				} else
+                    ++i;
+			} 
 		}
 
 		//updates the window elements when DPI is changed.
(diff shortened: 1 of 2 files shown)


ventsin:
For the compilation error, I think #665 is probably good if the behavior of `ellipse` was satisfactory already.
By explicitly converting to `int`, the behavior will remain the same while getting rid of the errors and also the warnings.

Unless rounding is desirable.

qPCR4vir:
The code was introduced days ago. Is new. Where exactly in code the convertions from float to int are calculated may affect the quality of the ellipse and artifacts may arise. I have not checked that. I din't not run that code. It just compile.

ventsin:
> Where exactly in code the convertions from float to int are calculated may affect the quality of the ellipse and artifacts may arise.

Yes, but the behavior would be exactly the same as Jinhao's code because all that's being done is making an implicit conversion into an explicit one, satisfying the compiler.

However, only @cnjinhao can say exactly if rounding would be desirable since he wrote it :laughing:
Hopefully he will shine a light on this.

cnjinhao:
Hi, the `ellipse` function is just a draft. I forgot to modify it before commit. After I check the #665 and #666(this).  I agree with @qPCR4vir 's fix, the implicit conversion is right behavior.

```c++
pixel(point(x - a, y), col);
//or
pixel({static_cast<int>(x - a), y}, col);

//for example.
//static_cast<int>(3 - 1.5) == 1
//3 - (int)1.5 == 2
```

cnjinhao:
Hi, the `ellipse` function is just a draft, I forget to check it before commit. I agree with @ventsin 's fix. The implicit conversion's problem in this issue is that int is implicitly converted to float. It causes following problem.

```
int x = 10;
float a = 1.5

int(x - a) = 8 //the distance from x is 2
int(x + a) = 11 // the distance from x is 1
```

The second issue, I think the call_safe_place is not a reentrant function, the after calling the function objects, the iterator `i` may be invalid.

PR #684, opened 2024-01-02
add single click event handler to listbox

crillion:
Hi, I've recently had the necessity, in a listbox, of managing not only the double click on row items, but also the single click. So I've tried to make a simple thing : I duplicated the code for the dbl_click handler in a click handler.
I've seen that it works : it correctly reacts to single clicks. To avoid code duplication I've put the common instructions in a central, private method.
One possible issue that remains (but I've no idea on how to fix it) is that now, whenever I double click, both events get called : double click and single click. Could single click be just called with single clicks ?


cnjinhao:
it's a little complicated. IMO, a proper way is to add a flag to indicate which handler should be invoked for listbox. The flag is a three-value flag: click/dbl_click/both

PR #732, opened 2026-03-08
Normvcr

normvcr:
Compiled for c++20, using the following environment variables:

export CXXFLAGS="-ggdb -std=c++20"
export CFLAGS="-ggdb"

Fixed some null pointer dereferences, and char/char8_t interfacing .


normvcr:
I inteded to merge into hotfix-1.7.4.  Will redo

PR #734, opened 2026-03-12
Fixed GCC compilation by adding <algorithm> to treebox.cpp

ventsin:
Not sure if the library is still under active development but I was patching up an old project and ran into this issue.
Thought I should submit a quick fix.


qPCR4vir:
Genial. Thank you!

